PR #1, opened 2020-03-01
Add 5 new effects.

Lucas7yoshi:
New effects:
- Jesus Take The Wheel: Player is moved to passenger seat and jesus is put into the driver seat and wanders around (at high speed)
- Spawn Bale Trailer: Kinda a reference to what someone was doing during FriendlyBaron's ChatCheat% run
- Teleport to random seat: Teleports you into a random seat of the current vehicle
- Spawn Random Vehicle now teleports the player into the drivers seat.
- 30MPH Speed Limit: self explanatory
- All Peds Ragdoll: also self explanatory

This is like my first time making a pull request so if i did something wrong sorry but hopefully you accept these new effects :)

### DIFF
--- a/ChaosMod/Effects/db/Vehs/Vehs30mphLimit.cpp
+++ b/ChaosMod/Effects/db/Vehs/Vehs30mphLimit.cpp
@@ -1,0 +1,22 @@
+#include <stdafx.h>
+
+static void OnStart()
+{
+	// Once the effect gets triggered
+}
+
+static void OnStop()
+{
+	// Once the effect stops
+}
+
+static void OnTick()
+{
+	auto ped = PLAYER_PED_ID();
+	if (IS_PED_IN_ANY_VEHICLE(ped, false))
+	{
+		SET_ENTITY_MAX_SPEED(GET_VEHICLE_PED_IS_IN(ped, false), 13.41);
+	}
+}
+
+static RegisterEffect registerEffect(EFFECT_VEH_SET_TOPSPEED_30MPH, OnStart, OnStop, OnTick);
--- a/ChaosMod/Effects/db/Vehs/Vehs30mphLimit.cpp
+++ b/ChaosMod/Effects/db/Vehs/Vehs30mphLimit.cpp
@@ -1,22 +1,12 @@
 #include <stdafx.h>
 
-static void OnStart()
-{
-	// Once the effect gets triggered
-}
-
-static void OnStop()
-{
-	// Once the effect stops
-}
-
 static void OnTick()
 {
 	auto ped = PLAYER_PED_ID();
 	if (IS_PED_IN_ANY_VEHICLE(ped, false))
 	{
-		SET_ENTITY_MAX_SPEED(GET_VEHICLE_PED_IS_IN(ped, false), 13.41);
+		SET_ENTITY_MAX_SPEED(GET_VEHICLE_PED_IS_IN(ped, false), 13.41); // 13.41 Meters Per Second = 30 MPH~
 	}
 }
 
-static RegisterEffect registerEffect(EFFECT_VEH_SET_TOPSPEED_30MPH, OnStart, OnStop, OnTick);
+static RegisterEffect registerEffect(EFFECT_VEH_SET_TOPSPEED_30MPH, nullptr, nullptr, OnTick);
--- a/ChaosMod/Effects/db/Vehs/Vehs30mphLimit.cpp
+++ b/ChaosMod/Effects/db/Vehs/Vehs30mphLimit.cpp
@@ -5,8 +5,17 @@ static void OnTick()
 	auto ped = PLAYER_PED_ID();
 	if (IS_PED_IN_ANY_VEHICLE(ped, false))
 	{
-		SET_ENTITY_MAX_SPEED(GET_VEHICLE_PED_IS_IN(ped, false), 13.41); // 13.41 Meters Per Second = 30 MPH~
+		_SET_VEHICLE_MAX_SPEED(GET_VEHICLE_PED_IS_IN(ped, false), 13.41); // 13.41 Meters Per Second = 30 MPH~
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pongo1231:
Would prefer for the spawn random vehicle to not set the player directly into it (as there are other effects that do that already). The new effects definitely look neat though :)

ChaosMod/Effects/db/Vehs/VehsJesusTakeTheWheel.cpp
```diff
@@ -0,0 +1,21 @@
...
+	auto playerPed = PLAYER_PED_ID();
+	if (IS_PED_IN_ANY_VEHICLE(playerPed, false))
+	{
+		auto veh = GET_VEHICLE_PED_IS_IN(playerPed, false);
+		auto seats = GET_VEHICLE_MODEL_NUMBER_OF_SEATS(GET_ENTITY_MODEL(veh));
+		if (seats >= 2)
+		{
+			SET_PED_INTO_VEHICLE(playerPed, veh, -2); // -1 = driver, -2 = First Free Passenger Seat
+			auto jesus = CREATE_PED_INSIDE_VEHICLE(veh, 4, modelHash, -1, true, false);
+			TASK_VEHICLE_DRIVE_WANDER(jesus, veh, 50, 1074528293); // 1074528293 = Rushed, 50 = 50 Meters per second = 111.847 MPH which i think is like the "target" speed.
```
pongo1231: Probably would be a great idea to also make the ped ignore whatever is happening around it (so it doesn't go into panic mode if there are gunshots being fired or a weapon aimed at it for instance). `SET_BLOCKING_OF_NON_TEMPORARY_EVENTS` would achieve that.

ChaosMod/Effects/db/Vehs/VehsSpawner.cpp
```diff
@@ -90,8 +90,20 @@ static void OnStartRandom()
 
 	if (!vehModels.empty())
 	{
-		CreateTempVehicleOnPlayerPos(vehModels[Random::GetRandomInt(0, vehModels.size() - 1)], GET_ENTITY_HEADING(PLAYER_PED_ID()));
+		auto vehHash = vehModels[Random::GetRandomInt(0, vehModels.size() - 1)];
+		Vehicle veh = CreateTempVehicleOnPlayerPos(vehHash, GET_ENTITY_HEADING(PLAYER_PED_ID()));
+		if (GET_VEHICLE_MODEL_NUMBER_OF_SEATS(vehHash) >= 1)
+		{
+			SET_PED_INTO_VEHICLE(PLAYER_PED_ID(), veh, -1);
+		}
```
pongo1231: Like already mentioned, would prefer the old behaviour

ConfigApp/Effects.cs
```diff
@@ -403,6 +410,9 @@ public enum EffectType
             {EffectType.EFFECT_LOW_POLY, new EffectInfo("Low Render Distance", EffectCategory.MISC, "world_lowpoly", true)},
             {EffectType.EFFECT_NEARBY_PEDS_OBLITERATE, new EffectInfo("Obliterate All Nearby Peds", EffectCategory.PEDS, "peds_obliterate")},
             {EffectType.EFFECT_VEHS_TRIGGER_ALARM, new EffectInfo("Alarmy Vehicles", EffectCategory.VEHICLE, "vehs_alarmloop", true)},
+            {EffectType.EFFECT_VEH_SET_RANDOM_SEAT, new EffectInfo("Teleport Into Random Seat", EffectCategory.VEHICLE, "veh_randomseat")},
+            {EffectType.EFFECT_VEH_SET_TOPSPEED_30MPH, new EffectInfo("30MPH Speed Limit", EffectCategory.VEHICLE, "veh_30mphlimit") },
```
pongo1231: Needs to be flagged as a timed effect here too (just append a true to the parameters).

Lucas7yoshi:
No problem. I'll make the changes once I get home, I also wanted to change the 30MPH speed limit to be a short duration instead of long, can I just use a nullptr for the things that it cannot work alongside option?

ChaosMod/Effects/db/Vehs/Vehs30mphLimit.cpp
```diff
@@ -0,0 +1,12 @@
...
+static void OnTick()
+{
+	auto ped = PLAYER_PED_ID();
+	if (IS_PED_IN_ANY_VEHICLE(ped, false))
+	{
+		SET_ENTITY_MAX_SPEED(GET_VEHICLE_PED_IS_IN(ped, false), 13.41); // 13.41 Meters Per Second = 30 MPH~
+	}
+}
+
+static RegisterEffect registerEffect(EFFECT_VEH_SET_TOPSPEED_30MPH, nullptr, nullptr, OnTick);
```
pongo1231: Should probably use `_SET_VEHICLE_MAX_SPEED` (`SET_ENTITY_MAX_SPEED` will limit the speed of things like falling down too). Also don't forget to reset the speed of all vehicles again on effect stop.

pongo1231:
> 
> 
> No problem. I'll make the changes once I get home, I also wanted to change the 30MPH speed limit to be a short duration instead of long, can I just use a nullptr for the things that it cannot work alongside option?

Just add a `{}` for that (look at how other short timed effects do it).

Lucas7yoshi:
does this look like it'll do the job of resetting the vehicles max speed? can't test ATM (doing this over phone)
`static void OnStop()
{
	// Cycles through all vehicles and sets their max speed to what it should be in case player changes vehicles mid effect
	for (auto veh : GetAllVehs())
	{
		_SET_VEHICLE_MAX_SPEED(veh, _GET_VEHICLE_MODEL_MAX_SPEED(GET_ENTITY_MODEL(veh)));
	}
}

`

pongo1231:
Yeah that should be fine.

Lucas7yoshi:
That should be everything you asked for now :)

30MPH thing works fine and jesus no longer reacts to guns

pongo1231:
Seems good :)

Issue #6, opened 2020-04-05
"Modders Stop Changing the Weather Plz" - epilepsy and motion sickness


brandishwar:
The effect has the potential to trigger photosensitive epilepsy or motion sickness in certain individuals due to how quickly it switches the weather.

I discovered this effect when DarkViperAU shared a [stream video](https://youtu.be/AC2aBo_1Fb4)  (at 34:37) to his YouTube channel. One commenter to the video said this effect gave him a headache and feeling of malaise. I also commented that it nearly triggered my motion sickness. I've also found another comment where someone pointed out they have photosensitive epilepsy and said the effect is a trigger.

Currently it appears the effect changes the weather at least once per second. At minimum it should be slowed down to once every two seconds or longer, or just remove entirely.

PR #7, opened 2020-04-05
Slowing down the random weather effect

brandishwar:
Issue #6 - slow down how often the weather is changed to avoid appearance of screen flicker


pongo1231:
2 seconds might be too long IMO. Would prefer it to be 1.5 secs.

brandishwar:
That sounds good, since that still shouldn't be too fast to be a problem. For me, at least. And it's still significantly better than how you initially had it.

PR #24, opened 2020-06-19
Add 3 new effects

yzimroni:
Hi, I've added 3 new effects:

1. No brakes - prevent the player from braking/descending
2. Weapon lock - prevent the player from changing weapons (except for when firing from a vehicle with a weapon that can't be fired from a vehicle and when the player runs out of ammo, in these cases the weapon changes automatically)
3. Fireworks - Fireworks show


pongo1231:
The `onStart` in MiscFireworks,cpp should start with a capitalized letter. Also ensure you put brackets in a new line so the formatting is on par with the codebase.

An easier and slightly more performant alternative for the `Weapon Lock` effect could be to hide the relevant hud component (WEAPON_WHEEL) via [HIDE_HUD_COMPONENT_THIS_FRAME](https://runtime.fivem.net/doc/natives/?_0x6806C51AD12B83B8) instead of disabling a lot of controls on every frame.

Otherwise this looks good and should be ready for merging if the changes mentioned above are done.

yzimroni:
I've fixed OnStart and put the brackets in new lines.
As for the weapon lock effect, I tried to use `HIDE_HUD_COMPONENT_THIS_FRAME`, but it just hides the weapon wheel, it still allows the player to change weapon (either from the invisible weapon wheel or using the next weapon button while in a vehicle).

pongo1231:
The "No HUD" effect completely disables weapon switching too (from what I've seen) so it has to be a combination of HUD components that need to be disabled for it to work properly. I'll probably look into that later.

Issue #30, opened 2020-07-11
[Suggestion] Fog of War


TheRealAyCe:
Shooting generates thick smoke. Don't know if peds care about being able to see. The idea is that you or they cannot continuously shoot from one place because the smoke makes vision impossible.

PR #50, opened 2020-07-22
Added "Fog of War" Effect

Last0xygen:
Filled Effect-Request: https://github.com/gta-chaos-mod/ChaosModV/issues/30
Hope this isn't too much 😄


pongo1231:
Make sure to capitalize the first letter of each word in the effect name in both the mod and the config utility (like every other effect in the mod).

Issue #39, opened 2020-07-11
[Suggestion] Oil spill


TheRealAyCe:
peds (and player? or cars only?) leave gas trails like when using the gas can

PR #51, opened 2020-07-22
Added Oil Leak Effect

Last0xygen:
Filled Effect-Request: #39


pongo1231:
Also including vehicles in the effect is probably a good idea ("Oil Leak" kind of implies that it affects vehicles mainly).

Last0xygen:
So should i only include vehicles then or just add them? :)

pongo1231:
Yup, just add them in addition to peds.

Last0xygen:
Should i move it to "Misc" then, as it is not "Peds only" anymore?

pongo1231:
A better implementation would probably be to loop through all peds which are not inside a vehicle and all vehicles and put them into a list (use a vector from the STL for that). Afterwards simply iterate through all the entities in the list and call `ADD_PETROL_DECAL` on them.

Also moving it into the Misc category is definitely not a bad idea. Make sure to also move and rename the source file correspondingly.

And I think "Oil Leaks" might be more descriptive than "Oil Leak" as it's multiple peds and vehicles leaking.

Last0xygen:
Changed everything as discussed.
I tested everything but there seem to be some areas, where the oil can't be deployed correctly, for example the area around Michael's home and the Pier area. Tried different things for some hours but can't get it to work over there 😞 
GET_GROUND_Z_FOR_3D_COORD returns true the petrol-decal doesn't get applied.

pongo1231:
There was some issue with oil leaks not showing up if there were too many existing entities, added a slight delay between every 25 entities which fixes it.

PR #53, opened 2020-07-23
Suicide Animation

Last0xygen:
Added the "Suicide Animation" to the Suicide-Effect if the player is not in a vehicle.
The player now stops, points a pistol at his head and fires before killing himself


pongo1231:
Make sure to wait for the anim dict to fully load before playing the animation.

PR #54, opened 2020-07-24
Added Effect Forcefield

Last0xygen:
Vehicles will be pushed away from the player


pongo1231:
Could be refactored into "Forcefield" if you also include peds and objects, which would make it less situational than only having it affect vehicles.

pongo1231:
Moved it into the Player category as I think it's a better fit in there.

PR #55, opened 2020-07-24
Added Effect These Boots Are Made For Running

Last0xygen:
Player keeps running, if he's on foot (and not aiming)


pongo1231:
The effect was still too "mild" in my opinion so I made some slight changes: in addition to `SIMULATE_PLAYER_INPUT_GAIT` it applies several controls usually mapped to W through `_SET_CONTROL_NORMAL` and disables all vehicle braking actions. I also renamed it to "Help My W Key Is Stuck" to fit the new changes. Right now it still isn't able to force the player to run in certain scenarios though (for instance during combat or while swimming), couldn't find a solution for that yet.

Issue #61, opened 2020-07-27
[Suggestion] Merryweather's here bro!


TheRealAyCe:
Spawn ane enemy helicopter that shoots you (with rockets?)

pongo1231:
Implemented in #63

TheRealAyCe:
@Last0xygen You've been going at these suggestions, great job! :+1:  Do you think it's possible to add the "Merryweather Fleet's here bro!" alert sound to the event, just like in the Impotent Rage one?

Last0xygen:
@TheRealAyCe I looked into that actually, but didn't find a way to trigger a campaign speech line

TheRealAyCe:
@Last0xygen Oh well, too bad.

TheRealAyCe:
@Last0xygen Right now the mercenaries seem to not be aggressive enough. Half the time it seems like you can't even notice them. Maybe you have an idea how to fix that? Are they not pursuing the player?

Last0xygen:
@TheRealAyCe i know. I guess the spawn too far away.
There’s already a fix for them to gespawn consistent (https://github.com/gta-chaos-mod/ChaosModV/pull/89).
I guess I have to adjust the random spawn location too, as it may be too far away 😅

TheRealAyCe:
@Last0xygen Oh, it's not merged in yet. What a pity. :( Hopefully the next Chaos stream will feature an updated version (which requires an updates version to be released...)

PR #63, opened 2020-07-27
Added Effect Merryweather Here

Last0xygen:
Spawns a buzzard with 4 merryweather guys in it, equiped with SMGs (i thought rocket launchers would be too overkill)
Requested in https://github.com/gta-chaos-mod/ChaosModV/issues/61


pongo1231:
Should persistently spawn mercenaries in both Buzzards and in Mesas from a distance away (check if they either died or are too far away, mark them as despawnable and spawn new ones if so), also would prefer a more generic effect name like "Mercenaries".

pongo1231:
In this case you should use the normal `CREATE_` natives instead of the custom pool entity functions. Those exist for entities which should persist longer than `SET_ENTITY_AS_NO_LONGER_NEEDED` allows them to but don't have their own management logic (which this effect does). I know it's not documented anywhere and I definitely should look into noting that down somewhere for everyone to see, I'm just saying that here now.

Also a distance check would definitely be a great addition to ensure there are always mercenaries close by if the player manages to escape them (a few hundred meters or so).

Last0xygen:
Changed it to use the native CREATE_ methods.
A distance-Check to 300 units has already been implemented

pongo1231:
Didn't see the existing distance check, my bad.

There has been a weird issue with neither enemies nor vehicles spawned by this not despawning at all (even after the effect stops), I've fixed this by setting them as mission entities just before marking them for despawning. I've replaced the existing `SET_X_AS_NO_LONGER_NEEDED` natives with custom functions which do this so this "fix" is applied to every effect.

PR #67, opened 2020-07-28
Add webpack-dev-server

byhemechi:


pongo1231:
Tagging @Elias-Graf

Elias-Graf:
Hey, cool that you took the time to add the `webpack-dev-server`. Would you mind updating the readme? So, remove the comment about it and state how to use it? Also, is there any particular reason you resolve the `.js` file extension?

byhemechi:
Alright, I've updated the README.

`js` files need to be resolved as `webpack-dev-server` tries to inject js files into the bundle for the hot reload and these obviously need to be resolved.

Elias-Graf:
Okay so maybe I'm mistaken, so please, educate me. 

I thought that resolve referees to the source files that should be compiled into js, not the files which should be served.

byhemechi:
Sorry for delay, it's required because webpack-dev-server injects itself into your scripts with a wepack require statement.

You get the same module resolution errors that importing any other js files does.

It's a really confusing error that i've been asked about many times (I'm a sysadmin)

Elias-Graf:
Okay if you say so... I just don't want anyone thinking they should add plain javascript files. Should be good to merge @pongo1231

PR #70, opened 2020-07-28
Added Effect "Fake Death"

Last0xygen:
Slows down the game and adds a fake wasted screen.
Don't know if this is a suitable Effect, found it funny. Had to add my own "stop", so it wouldn't take the whole 30 seconds to stop the effect


pongo1231:
Fixed a typo that prevented it from compiling successfully, made it a non-timed effect instead (as it doesn't need it and stops the "Fake Death" effect text from showing until the effect is over), shortened the second phase and moved it to the Player category in the config utility.

PR #71, opened 2020-07-29
Added Effect 'Replace Current Vehicle'

Last0xygen:
Replaces current vehicle with a random one


pongo1231:
Problem with this one is if it happens during a mission it won't retain the same handle as the old vehicle and break the mission in the same way "Remove Current Vehicle" does, which I would prefer to prevent. Spawning a new vehicle with the same handle will require research in the game code itself and won't be possible through natives.

Last0xygen:
yeah. Actually i took that into account but as the "remove current vehicle"-effect was "accepted" i thought this would be too.
Can be closed if not suitable :)

PR #72, opened 2020-07-31
Modified 'Forcefield'-Effect

Last0xygen:
* made peds ragdoll when in forcefield-range
* added lampposts and other props to be forced
* Fixed Filename and location (was still in the misc-category)


pongo1231:
The ability to push objects like street lights was also something I wanted to look into for the "Doomsday" effect, thanks for finding a way to do it.

Issue #31, opened 2020-07-11
[Suggestion] Loose gun (or Loose Cannon)


TheRealAyCe:
Weapons perma-shoot with infinite ammo. Either all peds or only player.
Maybe also lock current weapon, so that you can't switch to unarmed?

pongo1231:
If this were to be implemented, probably could be merged with the "No Weapon Switching" effect

pongo1231:
Implemented with #77

TheRealAyCe:
@pongo1231 @Last0xygen  This should probably switch to a gun if you are currently unarmed. The effect name also does not make the lack of weapon switching apparent, which may become less of a problem when the mod makes sure that the player always has something to shoot with.

PR #77, opened 2020-08-02
Added Effect 'Loose Trigger'

Last0xygen:
Requested here https://github.com/gta-chaos-mod/ChaosModV/issues/31
Also changed the PlayerWeaponLock-Effect to not show the Weapon-Wheel at all, not just ignore the selection.
Could be used to replace the PlayerWeaponLock-effect, if preferred


pongo1231:
Removed "No Weapon Switching" effect in favor of this one

Issue #79, opened 2020-08-04
My W Key is stuck does not walk forwards when aiming down the sights


TheRealAyCe:
Intended or not?

PR #82, opened 2020-08-04
Adjusted several effects

Last0xygen:
- Fixes https://github.com/gta-chaos-mod/ChaosModV/issues/79, so you can't aim or go to cover to escape the effect "Stuck W Key"
- Increased Recoil in "Heavy Recoil"
- Kills mercenaries if they're too far away. Watched DarkViper with that effect and they don't seem to respawn after mission failing and respawning
- Adjusted the Fake Death effect to show the current players animation
- "Loose Trigger" now affects all Peds (they keep shoot where their gun is pointing). Also the player can select weapons if a melee weapon is selected
- "Airstrike" now doesnt count as a shot by the player -> doesn't need to adjust the wanted level


TheRealAyCe:
Great job @Last0xygen. Glad I'm not the only Darkviper viewer here. :D Also noticed some of those things but I wanted to wait before reporting them. You fixed them preemptively, nice.

PR #93, opened 2020-08-07
Added new effect "Simon Says"

Slothersbee:
Simon Says -
Gives you an rule that if broken you get blown up :)
E.g. Simon Says: Move backwards = if you let go of back you will be blown up
E.g. Simon Says: Don't move left = if you move left (press a) you will be blown up

### DIFF
--- a/ChaosMod/ChaosMod.vcxproj
+++ b/ChaosMod/ChaosMod.vcxproj
@@ -125,6 +125,7 @@
     <ClCompile Include="Effects\db\Player\PlayerHeavyRecoil.cpp" />
     <ClCompile Include="Effects\db\Player\PlayerKeepRunning.cpp" />
     <ClCompile Include="Effects\db\Player\PlayerLooseTrigger.cpp" />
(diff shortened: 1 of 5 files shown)


LennartOschika:
What if the Effect was called "Simeon" says instead of Simon says?

Slothersbee:
> What if the Effect was called "Simeon" says instead of Simon says?

YESSSSS, I love that, haha thanks for the suggestion, I'll drop that in.

TheRealAyCe:
@Slothersbee This is incompatible with Need for Speed and Fake Death due to MP_BIG_MESSAGE_FREEMODE, right? Has that been taken care of?

Slothersbee:
> @Slothersbee This is incompatible with Need for Speed and Fake Death due to MP_BIG_MESSAGE_FREEMODE, right? Has that been taken care of?

The popup only come up for the first two seconds - I've tried it with both fake death and need for speed and it doesn't cause any issues 👍

PR #94, opened 2020-08-07
Added new effect "Poof"

Slothersbee:
Poof - Obliterates any ped/vehicle that you aim at.

Could possibly do with a name change? What do you think?


Slothersbee:
Not sure what's going on with the EffectsInfo.h file change? Something to do with different line endings maybe? I'll try and fix it.

PR #102, opened 2020-08-21
Added Effect 'Near Sighted'

Last0xygen:
Timecycle-Modifier, that only shows entities within a certain (small) range


pongo1231:
I fear that might be too similar to existing effects like Extreme Fog and Low Render Distance so I'll have to decline this one unfortunately.

PR #103, opened 2020-08-21
Added Effect 'I Need Glasses'

Last0xygen:
Blurs the vision


pongo1231:
This one will probably replace "Blind" later.

PR #115, opened 2020-09-04
Added Effect 'To Drive Or Not To Drive'

Last0xygen:
Burst all tires for 750ms every 1-5 seconds


pongo1231:
Practically the same as "Random Tire Popping", makes driving harder by repeatedly popping tires.

PR #116, opened 2020-09-05
Adjusted 'Set Everyone into random vehicles'

Last0xygen:
Trying to prevent the game from crashing.
Only allow setting into valid vehicles are not underground


pongo1231:
This should still include the player, regardless of if they are in a vehicle or not. Also this would also exclude any broken vehicles which is part of the fun. It should instead exclude whales (spawned by "Whale Rain") which might be one potential cause for crashes caused by the effect.

Last0xygen:
This does include the player, it just wont port any ped, if they are not in any vehicle (unnecessary wait-delays)
I didn't really think about animals being peds, you're right. As i don't how to check every Ped, if they're able to be placed inside a vehicle, i would only allow human peds to be ported

PR #120, opened 2020-09-07
Added Effect 'I'm So Tired'

Last0xygen:
Fades the screen out, everytime slower than before


pongo1231:
Hehe I like this one :)

PR #128, opened 2020-09-27
Added Effect 'KickFlip'

Last0xygen:
Flips the Car or the player


pongo1231:
Since it also affects the player on foot it should be in the "Player" sub-category.

Last0xygen:
Done

DEATHCON1:
A possible rename for this suggestion could be: 'Do A Barrel Roll!'

Issue #106, opened 2020-08-24
Lock vehicle camera, oil slicks and a bullet


diamondo25:
This combination seems to make your camera really shaky when going left or right, almost nauseating.

Last0xygen:
The problem here, is that the 'Lock Camera' effect has to be executed every frame to function properly.
The 'Oil Trails' (and other effects) have "WAIT(0)" inside their OnTick methods. As all effects are called synchronously this delays the execution of the 'Lock Camera' Tick-Method.
I don't know what the best solution would be. 
Executing them in seperate threads may help but i don't want to start that without getting feedback.
What do you think, @pongo1231 ?

pongo1231:
@Last0xygen Could definitely be one way to avoid the whole script from pausing when invoking `WAIT`.

PR #131, opened 2020-09-29
Added Multithreading

Last0xygen:
This commit adds multithreading to the effectdispatcher.
This should fix https://github.com/gta-chaos-mod/ChaosModV/issues/106 but also increase overall performance, as it won't block the main thread anymore
Although i did test a lot of effects and effect combinations this should be tested thoroughly :)
Also, maybe reconsider if the "WAIT(0)"-calls  are now even necessary anymore in all effects :)


TheRealAyCe:
@Last0xygen Just asking, how does the threadingmodel for GTA5 mods like this one look? What is the wait(0) for? I would imagine you can't just spawn stuff without stopping the GTA game thread for a while, reserving your turn to modify the memory, so to speak. Do you know about that?

Last0xygen:
Short answer:
to give the game time to deal with and execute  your changes

It appears like so:
You can have multiple background Threads creating stuff -> the effects 
To actually handle them they have to be rendered and executed in the games „main thread“/„rendercycle“
This is where the „WAIT(0)“ comes in. It pauses the current thread to give the main Thread time to execute its pending work.
This is required for several reasons:
- There’s a limit for objects being spawned at a time. If you try to spawn too many objects without giving the game time to handle them they will be ignored 
- after spawning a vehicle for example you have to wait one cycle (WAIT(0)) for the game to correctly handle/spawn that vehicle. Only after that you‘ll be able to for example place a ped into it 
- freezing: this was the problem before this PR. There’s one thread handling all effects -> if one effect needs to pause for some time, all other effects are delayed which may cause weird behavior if some effects don’t get executed every Frame

But to your last question: you‘re not blocking the game, Threads are running asynchronous

I hope I (at least partly) answered your question

TheRealAyCe:
>To actually handle them they have to be rendered and executed in the games „main thread“/„rendercycle“

But how can the main thread know what it needs to do? The code suggests that you just create the object like here:
`Ped whale = CREATE_PED(28, WHALE_MODEL, spawnPos.x, spawnPos.y, spawnPos.z, g_random.GetRandomInt(0, 359), true, false);`
This needs to create the object in memory and activate it for the game somehow, no? I'm worried that this would need to make sure that the game's main thread is not currently accessing the list of object too, otherwise bad things could happen. I'm used to having a "SynchronizationContext" or "Dispatcher", where I can schedule delegates to run on some thread, so that worker threads can communicate their results (back) to the main thread.

>freezing: this was the problem before this PR. There’s one thread handling all effects -> if one effect needs to pause for some time, all other effects are delayed which may cause weird behavior if some effects don’t get executed every Frame

It seems weird to me that you would need to create a C++ thread and then pause the entire thread just because you are waiting for something. In C# you would do this by `await`ing, so that your thread is not actually paused, but other code can run in the meantime. If you are in a worker thread, that thread may do other work until the condition that is `await`ed is resolved. The main thread may process other pending messages and will eventually get back to the code that `await`ed to continue execution, once the condition is met. I thought C++ introduced something like that too recently, hence why I find it strange to need multiple threads.

>But to your last question: you‘re not blocking the game, Threads are running asynchronous

I was less asking about blocking the game from a performance view, more as a necessity. How else would you ensure that the game itself does not try to modify the same memory as you do at the same time? This could lead to all sorts of glitches, crashes, bugs and so on. How is this prevented? That's what I thought the WAIT(0) is for, wait until the main thread is sleeping, and prevent it from waking up again until the mod does its memory modifications.

Because that's what the mod does, right? It does not tell the game to spawn something through an API or a hook, but it just calls the methods in memory directly, correct? (Hence the `Hash` stuff for searching for models in memory?) From the thread that the mod spawned, not the main thread of the game. But it accesses the same memory.

Thanks for explaining this stuff to me! :)

Last0xygen:
I sure hope I’m not talking bullshit here so please correct me if I’m wrong ;P

First I should revoke my previous statement. The „WAIT“-functions does not invoke a „main thread“ update per se, but gives its engine time to compute pending actions as it sleeps it's thread 😅 

Also:
Please do not take this PR as a final version of how it has to be implemented. I'm not a c++ developer (but swift). I built something which does work for me but i may have missed something, maybe there is a way easier method of doing this, I'm just trying to take first step into doing this as i think this would greatly improve the performance of this mod.

> But how can the main thread know what it needs to do? The code suggests that you just create the object like here:
> Ped whale = CREATE_PED(28, WHALE_MODEL, spawnPos.x, spawnPos.y, spawnPos.z, g_random.GetRandomInt(0, 359), true, false);
> This needs to create the object in memory and activate it for the game somehow, no? 
>
Yes, this calls the scripting-engine which will create an object and return its pointer.
This should be able handle asynchronous calls, otherwise the game itself (or better it’s scripts) couldn’t be processed asynchronous themselves.

>I'm worried that this would need to make sure that the game's main thread is not currently accessing the list of object too, otherwise bad things could happen.
> 
This already is a problem. If you have a loop through all Vehicles for example and have a WAIT in there to apply several things and the game decides to delete said vehicle while said loop, this may lead to crashes/bugs/unexpected behaviour.
This would be a problem, if the natives are directly accessing game memory. The game engine handles those lists though, so hopefully there's no real need for native-Calls to be more careful than with the current synchronous approach. 
If there is a conflict or error case here, we should implement some sort of pool every thread can acces (synchronous) to eliminate those errors.
Memory updates should probably be made synchronous if they edit the same lists, yes.
Any static lists created by the mod possibly should therefore be synchonized. This could be a problem now, but something that should be preventable.
All effects running synchronous will 100% be a problem in the long run though (see https://github.com/gta-chaos-mod/ChaosModV/issues/106)

> It seems weird to me that you would need to create a C++ thread and then pause the entire thread just because you are waiting for something. In C# you would do this by `await`ing, so that your thread is not actually paused, but other code can run in the meantime. If you are in a worker thread, that thread may do other work until the condition that is `await`ed is resolved. The main thread may process other pending messages and will eventually get back to the code that `await`ed to continue execution, once the condition is met. I thought C++ introduced something like that too recently, hence why I find it strange to need multiple threads.
> 
I'm sorry for my poorly chosen words, i didn't know on what level i could communicate :D 
I'm not an expert in C++ so i don't know if anything like this has been added recently 😅
But i do know i failed to make a working version of creating threads after threads for each new effect and properly stopping them again. I also found it to be a "cleaner" way to let each thread work for itself, just being fed an effect.
But i'll look again into creating new threads if needed and not creating a bunch on start :D 

> I was less asking about blocking the game from a performance view, more as a necessity. How else would you ensure that the game itself does not try to modify the same memory as you do at the same time? This could lead to all sorts of glitches, crashes, bugs and so on. How is this prevented? That's what I thought the WAIT(0) is for, wait until the main thread is sleeping, and prevent it from waking up again until the mod does its memory modifications.
>

Uhm... well.. the game freezes while the thread's not sleeping, so that may be so 😆 🤦‍♂️  
So the game waits for mod to "finish" and continues while the mod is "sleeping"/WAITing

> Because that's what the mod does, right? It does not tell the game to spawn something through an API or a hook, but it just calls the methods in memory directly, correct? (Hence the Hash stuff for searching for models in memory?) From the thread that the mod spawned, not the main thread of the game. But it accesses the same memory.
> 
It does call engine-methods in memory directly, correct.

> Thanks for explaining this stuff to me! :)

Please, PLEASE, don't take my word for the real deal. I'm just talking about what i found out, maybe its all bullshit I'm talking 🤷‍♂️

TheRealAyCe:
>Also:
Please do not take this PR as a final version of how it has to be implemented. I'm not a c++ developer (but swift). I built something which does work for me but i may have missed something, maybe there is a way easier method of doing this, I'm just trying to take first step into doing this as i think this would greatly improve the performance of this mod.

Hey, no worries, I'm just trying to understand/learn how this mod operates since I've become interested in its inner workings! :D

>First I should revoke my previous statement. The „WAIT“-functions does not invoke a „main thread“ update per se, but gives its engine time to compute pending actions as it sleeps it's thread sweat_smile

>Uhm... well.. the game freezes while the thread's not sleeping, so that may be so laughing man_facepalming
So the game waits for mod to "finish" and continues while the mod is "sleeping"/WAITing

This is exactly what I thought needs to happen. If you have a `while(1){}` in the mod code, the game should freeze completely, that would indicate that the synchronization between mod and game works as I think it should. So WAIT will let the game run for a frame and then block it again to let the mod do things, which are hopefully so fast that no delay is noticeable ingame?

>Yes, this calls the scripting-engine which will create an object and return its pointer.
This should be able handle asynchronous calls, otherwise the game itself (or better it’s scripts) couldn’t be processed asynchronous themselves.

I don't know what scripting language this uses, but I think those scripts would run inside a VM that would not actually have Threads, but something more comparable to Tasks(/Fibers?). So if a script says "Sleep for 1 second", the game would not have an actual thread sleeping for one second, but store that script execution stack somewhere and start a timer to get back to it again after 1 second, when no other script actively runs, because all scripts need to run on the main thread, due to synchronization. But that's just how I'd imagine it would be done since that's how other scripting languages behave.

>This already is a problem. If you have a loop through all Vehicles for example and have a WAIT in there to apply several things and the game decides to delete said vehicle while said loop, this may lead to crashes/bugs/unexpected behaviour.
This would be a problem, if the natives are directly accessing game memory. The game engine handles those lists though, so hopefully there's no real need for native-Calls to be more careful than with the current synchronous approach.

If you would call WAIT inside the loop, I think this behaviour would be normal, since you actually wait for the game to run a frame, which most likely will change the internals of the iteration you are in. But if you do something inside that loop that just takes a little while longer (something similar to the `while(1){}` above), I would expect the game to lag, but not run a frame during our iteration of the vehicle list, so unless we are doing a WAIT, there should not be a danger of crashes here, right?

>All effects running synchronous will 100% be a problem in the long run though (see #106)

I'm not sure if it's a good idea to have multiple threads all use WAIT(0). Maybe the modding framework takes care of that and registers all mod threads that are waiting for an iteration of the game loop, but maybe this works differently than I think. If it does work like that, it should be fine.

If mod threads A and B are WAITing at the same time, will the game
- process a frame then get back to A, process another one and get back to B?
- process a frame then get back to both? (but how does it know that it needs to wait for both to finish their memory changes?)

I guess what I would expect is to have a sort of critical section in the mod code, inside of which you would call the game's functions. This critical section would ensure that your code blocks the game from running another frame until you are done. This could then also offer a WAIT function to temporarily suspend the critical section until a frame has been processed. Maybe the whole mod code already runs inside such a critical section? If that was the case, then I don't understand how the 2 mod threads would ensure not to modify the memory at the same time.

>Please, PLEASE, don't take my word for the real deal. I'm just talking about what i found out, maybe its all bullshit I'm talking man_shrugging

Maybe it's all bullshit I'm worried about too. I'm just wondering if the mod crashes are due to programming error in the effects, or due to the whole mod framework thing not being synchronized to the game properly so that mod and game occasionally access the memory at the same time.

pongo1231:
While the suggested way in the PR will work, I think it's better to create the threads on demand (and stop them as soon as an effect ends) instead of having a fixed amount of threads running constantly in the background. This will improve performance and remove the limit of timed effects running at once this approach would impose.

pongo1231:
> Just asking, how does the threadingmodel for GTA5 mods like this one look like?

Currently there are only 3 threads (2 fibers and 1 actual thread handling keyboard input) which are registered once the .dll (.asi scripts are just a renamed .dll) is attached to the running game process by the asi loader.

The "main" thread which handles the drawing of the timer bar, effect texts, debug menu and several other non-effect specific things.

There's also an additional thread handling all the effects themselves, twitch voting and all of the keyboard shortcuts.

And lastly there's an actual thread used for handling keyboard input. Since it's a genuine thread running "natives" (game functions) in there is not advised and will very likely cause a crash.

Currently an issue the mod has is that executing scripthook's `WAIT` function pauses the execution of the current fiber to allow the game to handle other tasks (either to prevent the game from completely locking up in the case of waiting for a model to load which obviously needs to be done by the game itself for instance or to prevent a game crash from doing too many changes (like giving a dozen of peds tasks) in 1 frame). Since all the effects share the same thread this can cause effects which require a native to run at every frame to not function properly.

This fixes the issue by splitting the effects up into their own threads.

> What is the wait(0) for?

Like said above, this pauses the execution of the current fiber for a specific amount of milliseconds (specified in the argument). `0` essentially means "resume as soon as possible".

> I would imagine you can't just spawn stuff without stopping the GTA game thread for a while, reserving your turn to modify the memory, so to speak.

Correct, you take your "turn" to invoke natives safely and ideally you want to either keep it short or amortize your code to not noticeably affect game performance.

TheRealAyCe:
>Currently there are only 3 threads (2 fibers and 1 actual thread handling keyboard input) which are registered once the .dll (.asi scripts are just a renamed .dll) is attached to the running game process by the asi loader.

Ah, so GTA5 knows how to load ASI files? Or is Scripthook a "meta-mod" to inject code?

>Currently an issue the mod has is that executing scripthook's WAIT function pauses the execution of the current fiber to allow the game to handle other tasks (either to prevent the game from completely locking up in the case of waiting for a model to load which obviously needs to be done by the game itself for instance or to prevent a game crash from doing too many changes (like giving a dozen of peds tasks) in 1 frame). Since all the effects share the same thread this can cause effects which require a native to run at every frame to not function properly.

This is exactly how I was thinking the system would need to operate in order for it to work properly.

>This fixes the issue by splitting the effects up into their own threads.

Is every "thread" (actually a "fiber", since it's taken care of by Scripthook, right?) synchronized in this fashion? So will they all wait for one game frame and then run one after the other?

Amazing, thank you very much for the clarification! :D

pongo1231:
> Ah, so GTA5 knows how to load ASI files? Or is Scripthook a "meta-mod" to inject code?

asiloader is the one handling ASI files. It gets loaded by the game by naming itself after one of the libraries which get automatically loaded (in this case dinput8.dll) and simply loads all ASI files through (I assume) `LoadLibrary`.

> Is every "thread" (actually a "fiber", since it's taken care of by Scripthook, right?) synchronized in this fashion? So will they all wait for one game frame and then run one after the other?

All fibers will run until they invoke scripthook's `WAIT` function, which pauses the current fiber and gives the other fibers and the game itself time to run. This function has to eventually be called by every fiber running a while true loop since the game would otherwise freeze.

TheRealAyCe:
>asiloader is the one handling ASI files. It gets loaded by the game by naming itself after one of the libraries which get automatically loaded (in this case dinput8.dll) and simply loads all ASI files through (I assume) LoadLibrary.

Neat!

>All fibers will run until they invoke scripthook's WAIT function, which pauses the current fiber and gives the other fibers and the game itself time to run. This function has to eventually be called by every fiber running a while true loop since the game would otherwise freeze.

So those fibers are not real threads but started by the GTA5 main thread itself? Or does every fiber have to start with a WAIT(0)?

pongo1231:
> So those fibers are not real threads but started by the GTA5 main thread itself? Or does every fiber have to start with a WAIT(0)?

Yes, they all share the same thread. Fibers containing an infinite loop need a `WAIT(0)` call somewhere to avoid starving the main thread and causing a game freeze, otherwise it's usually not required.

TheRealAyCe:
>Yes, they all share the same thread. Fibers containing an infinite loop need a WAIT(0) call somewhere to avoid starving the main thread and causing a game freeze, otherwise it's usually not required.

Do you know where I can learn more about this exact topic? I wonder how the game is able to "cut" the method call so that WAIT(0) is turned into "wait until the main thread resumes the fiber", similar to C#'s `async/await`. I would assume that needs compiler support or something.

pongo1231:
> Do you know where I can learn more about this exact topic? I wonder how the game is able to "cut" the method call so that WAIT(0) is turned into "wait until the main thread resumes the fiber", similar to C#'s async/await. I would assume that needs compiler support or something.

ScriptHookV actually overrides that function (the game's `WAIT` native only works for it's own game scripts). I don't know what it's doing exactly but I assume it makes use of win32's fiber API to jump between all of the fibers.

TheRealAyCe:
>ScriptHookV actually overrides that function (the game's WAIT native only works for it's own game scripts). I don't know what it's doing exactly but I assume it makes use of win32's fiber API to jump between all of the fibers.

Oh, Win32 API natively works with Fibers. Nice!

Last0xygen:
Can't seem to register an additional thread after the dllmain-method :/ 
Always crashing with "FATA: Trying to register additional thread 0x.... to unk script" :|
Could someone else look at it too? 😅 (looking at you, @pongo1231  and @drunderscore  😉 )

drunderscore:
hm... I'm having some a slightly different problem than that.
I can register an additional thread inside of dllmain, and have it execute. I can also register additional threads outside of the dllmain, but they will not execute once registered. See the output below inside of my test mod. (I tested with ChaosMod afterwards and got similar results)
[output](https://pastebin.com/u8xN3MQ0)

drunderscore:
After reading through this ~thread~ issue (poor word choice), I've come to my own conclusion that registering scripts on-demand for dispatching each effect is not desirable (or possible).

Here are my suggestions, which I'm sure have already been mentioned, but for my own sake:
- Run all effects synchronously on the same thread (aka how it is now):
The primary change would be our usage of `WAIT`. The only wait should be inside the effect main loop -- no effect should use it. We can circumvent it's usage by using the following paradigm:

```cpp
static Hash bullet_model = 0;
static Vehicle veh = 0;

void OnStart() // No signature change
{
    bullet_model = MISC::GET_HASH_KEY("BULLET");
    // Do things that need to be done on startup in startup
    STREAMING::REQUEST_MODEL(bullet_model);
    // Normally, we would use a busy loop to check if the model was loaded
    // with STREAMING::HAS_MODEL_LOADED and WAIT. Instead, we move
    // this check to the OnTick, and if it's not loaded, we simply return early.
    // Execution remains on our thread for other effects to continue like normal,
    // and this effect will be able to wait for something to load.
    // This approach perfectly solves the #106 problem.
}

void OnTick() // No signature change
{
    if(!STREAMING::HAS_MODEL_LOADED(bullet_model))
        return;
    veh = VEHICLE::CREATE_VEHICLE(bullet_model, ...); // Line unfinished
    STREAMING::SET_MODEL_AS_NO_LONGER_NEEDED(bullet_model);
}
```

Taking a look at Oil Trails, the wait is used to allow more decals to be placed than the game will allow at once. There may or may not be a solution to that -- I personally see it as a flaw trying to spawn so many.

- Register `x` amount of additional threads of startup.
These are allocated for effects, and each effect has a dedicated thread.
This seems like a good simple solution, however I fear a few problems:

  - Creating entities/blips/ropes on one thread will prevent the primary effect dispatcher thread from deleting them. (we would have to ensure everything is cleaned up from it's respective thread)
  - We have an even harder limit on effects
  - If the `OnEnd` has a wait, the next effect won't start immediately, and will cause more chaos<sub><sub>lol</sub></sub> than need be.


Certainly looking for comments opinions and corrections!

Last0xygen:
> * The primary change would be our usage of WAIT. The only wait should be inside the effect main loop -- no effect should use it.
Considering this would require to remove any "WAIT("-usage this would result in all effects being broken, that 
* try to spawn vehicles and put peds in them (as it would sometimes require some WAIT to spawn the vehicle
* Loop through all vehicles, as this requires some loop to not stutter
* multiple effects loading the same model this way would result in effect1.onStop removing the model and effect2.onTick not having the model anymore

> This seems like a good simple solution, however I fear a few problems:
>       
> * Creating entities/blips/ropes on one thread will prevent the primary effect dispatcher thread from deleting them. (we would have to ensure everything is cleaned up from it's respective thread)
Can't you remove entities from another thread? Like queueing via any of the PoolSpawner-functions 🤔 
> * We have an even harder limit on effects
What limit are you talking about? Simultaneous effects?
> * If the `OnEnd` has a wait, the next effect won't start immediately, and will cause more chaoslol than need be.
Only if you run out of effect-threads 😅 

> Edit 1:
> Here's a similar issue that occurs due to `WAIT`s in scripts
> image
Yeah, this can go on and on...
Any visual effect is broken if theres another effect using "WAIT(0)" in its tick (any looping through vehicles effect)
this results in flickering for effects that require execution on every frame (OnDemandTV, DVD-Screensaver, I'm so tired...)

pongo1231:
I'll merge this for now; the limit of 20 simultaneous effects should be high enough and worth it for the amount of issues it resolves. Might figure out a way to avoid a hard limit later, though not being able to register additional threads inside of scripthook threads (which I'm guessing the issue is) will make this harder.

PR #132, opened 2020-09-29
Adjusted 'Fake Death'

Last0xygen:
It now plays the suicide animation or adds an explosion if in vehicle


pongo1231:
Maybe make it random (either suicide / vehicle explosion OR nothing) to add variety to it.

TheRealAyCe:
Can this actually cause a mission fail though due to breaking the doors? ;)

Issue #104, opened 2020-08-21
[Suggestion] Spawn Balla Squad


JustTunerYT:
4 times 'Spawn Random Enemy' but model ID is ballas. Each time effect occur, it spawns random 4 Balla guys. @pongo1231 i think you should pick the choice. Either they Spawn in the Car or on the ground (Like on surface, in vehicle or in aircraft with player)
Balla squad have Micro-UZI, just like in their 'hood.

PR #133, opened 2020-09-30
Added Effect 'Balla Squad'

Last0xygen:
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/104


ChaosMod/Effects/db/Peds/PedsSpawnBallaSquad.cpp
```diff
@@ -0,0 +1,49 @@
+/*
+	Effect by Last0xygen
+*/
+
+#include <stdafx.h>
+
+static void OnStart()
+{
+	std::vector<int> ballaNames = { -198252413, 588969535, 361513884, -1492432238, -1410400252, 599294057 };
```
pongo1231, line 9: Not really a big deal in this case, but this should be `static const` so you don't create a new list every time the effect starts

PR #136, opened 2020-10-03
Adjusted Timecycle Effects

Last0xygen:
Fades in now instead of instant changes


Last0xygen:
This would need to be implemented for https://github.com/gta-chaos-mod/ChaosModV/pull/103 too

PR #138, opened 2020-10-16
Implement Turn Turtle (player_flip_camera)

### LINKED ISSUE #81: [Suggestion] Grand Theft Australia
Flip the player camera so that the world looks 180° rotated.

Alternative names:
GTAus
Is This What Australia Looks Like?
GTA Adelaide
Flip Camera (booooring)

drunderscore:
This will create a new rendering camera, make it active, and render it.
Each tick it will follow the gameplay camera, except have it's roll set
to 180 degrees.

Currently, the biggest issue is that the camera will not go back to
normal once the effect ends. In my own mod, it works perfectly, but
inside of chaos something is different that I've yet to figure out. Any
input on this would be much appreciated.

Resolves #81


drunderscore:
Perfect, all ready for merge!

PR #140, opened 2020-10-16
Implement Peds Drive Backwards

drunderscore:
When activated, all peds in vehicles will have their task driving style set to 1024,
which will make them perform their vehicle task in reverse gear.


pongo1231:
Would be better suited as a timed effect (with normal duration).

drunderscore:
Implemented suggestion changes and re-based

Issue #144, opened 2020-10-20
Can Peds be made to walk on water?


TheRealAyCe:
If that was possible, maybe upgrade Jesus?

PR #148, opened 2020-10-24
Added Effect 'Walk On Water'

Last0xygen:
Enables walking and driving on water (player only)
(Possible) Fix https://github.com/gta-chaos-mod/ChaosModV/issues/144


pongo1231:
Doesn't seem to work for me, even if I try to teleport above the ocean my player just falls into the water as usual.

Last0xygen:
Hey, maybe i should wait for the model to load before using it 😂

pongo1231:
Works now, though I noticed if the player is already swimming in water when this effect triggers it'll simply push them under water and spawn the object above them. Checking if the player is already in water and teleporting them above it will fix that.

Last0xygen:
Good finding, fixed it

pongo1231:
Perfect

PR #160, opened 2020-11-04
Added Effect 'Rapid Fire' and a bunch of Vector-Methods

Last0xygen:
Added a bunch of Vector3-Helper Methods and a function to get the current GameplayCamera and relative offset.
Ported a lot of them from scripthookvdotnet


pongo1231:
Nice job with the helper functions, will definitely come in handy in the future

PR #163, opened 2020-11-10
New Effects - Dice Roll

vladig98:
Two new effects that simulate a die roll.

I decided to make two dice, one bad and one good.

The bad die has the following outcomes:
1 - spawns the slowest vehicle in the game
2 - removes all weapons
3 - removes your money
4 - teleports you to the edge of the map
5 - kills you
6 - 5 stars wanted level

The good die has the following outcomes (counterparts):
1 - spawns the fastest vehicle
2 - gives all weapons
3 - gives max amount of money
4 - teleports to waypoint
5 - max health and armor
6 - clears wanted level

### DIFF
--- a/ChaosMod/Effects/EffectsInfo.h
+++ b/ChaosMod/Effects/EffectsInfo.h
@@ -226,6 +226,8 @@ enum EffectType
 	EFFECT_FLIP_CAMERA,
 	EFFECT_PLAYER_ON_DEMAND_CARTOON,
 	EFFECT_PEDS_DRIVE_BACKWARDS,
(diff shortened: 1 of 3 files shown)


pongo1231:
Seems like a good idea, however I'd do a few changes: instead of 2 "Bad RNG Die" and "Good RNG Die" effects, simply create one "Dice Roll" effect which has the chance to do any of the good or bad actions you implemented. Also in my opinion the "You rolled X" message doesn't really add anything to the effect and should rather be completely removed. Otherwise a good addition to the mod. :)

PR #164, opened 2020-11-12
Added effect Jesus Army

ProfessorBiddle:
Spawns 20 very displeased Jesus' with Baseball Bats


ProfessorBiddle:
Replaced with #228

PR #165, opened 2020-11-12
Added Effect Random Traffic

ProfessorBiddle:
Removes peds from vehicles, Randomizes all vehicles on the map except mission vehicles (hopefully), and places the peds back in them.

Limited to short effect because giving peds GTA online vehicles can cause bad things to happen very quickly


ChaosMod/Effects/db/Vehs/VehsRandomTraffic.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+			//spawn random vehicle in place of old vehicle
+			Vector3 prevPos = GET_ENTITY_COORDS(veh, false);
+			float prevRot = GET_ENTITY_HEADING(veh);
+			SET_ENTITY_AS_MISSION_ENTITY(veh, true, true);
+			DELETE_VEHICLE(&veh);
+			static std::vector<Hash> vehModels = Memory::GetAllVehModels();
+			if (!vehModels.empty())
+			{
+				Vehicle veh = CreatePoolVehicle(vehModels[g_random.GetRandomInt(0, vehModels.size() - 1)], prevPos.x, prevPos.y, prevPos.z, prevRot);
```
pongo1231, line 29: Use `CreateTempVehicle` if possible to avoid despawning other entities added to the pool by the mod (since it'll easily hit the 30 entities limit).

PR #166, opened 2020-11-12
Added Effect Gravity Field

ProfessorBiddle:
Adds player centered force on all objects proportional to distance, makes for some really funny cutscenes and usually immediate death.


ChaosMod/Effects/db/Player/PlayerHasGravity.cpp
```diff
@@ -0,0 +1,60 @@
...
+*/
+
+#include <stdafx.h>
+
+static void OnTick()
+{
+	Ped playerPed = PLAYER_PED_ID();
+	std::vector<Entity> entities;
+	//make player not die instantly
+	SET_PLAYER_INVINCIBLE(playerPed, true);
```
pongo1231, line 12: First arg requires the player (`PLAYER_ID()`) to actually work, not the player ped

PR #167, opened 2020-11-12
Added effect Bus Bois

ProfessorBiddle:
Similar to Scooter Brothers but bigger and more explodey.


ChaosMod/Effects/EffectsInfo.h
```diff
@@ -466,4 +467,5 @@ const std::map<EffectType, EffectInfo> g_effectsMap =
 	{EFFECT_FLIP_CAMERA, {"Turn Turtle", "player_flip_camera", true}},
 	{EFFECT_PLAYER_ON_DEMAND_CARTOON, {"On-Demand TV", "player_on_demand_cartoon", true}},
 	{EFFECT_PEDS_DRIVE_BACKWARDS, {"Peds Drive Backwards", "peds_drive_backwards", true}},
+	{EFFECT_PEDS_BUSBOIS, {"Bus Bois", "peds_busbois", false, {}, false}},
```
pongo1231, line 470: Just as a hint for next time: the default values for those arguments are already false, so no need to explicitly add them.

PR #168, opened 2020-11-12
Added Effect Spinning Peds

ProfessorBiddle:
Same type of effect as slidy peds, makes peds spin really fast around all axis'.


pongo1231:
Looks good, though as mentioned in another effect you shouldn't forget to also add your credits to the effect files (assuming you want to include them of course).

PR #171, opened 2020-11-12
Added Effect Schizophrenia

ProfessorBiddle:
Makes everything but the player and the vehicle the player is in invisible, so that you hear sounds and voices and run into things you can't see


pongo1231:
I feel like it's too similar to "Where Did Everything Go?" and "Low Render Distance" as it accomplishes the same as those effects (minus the world still being rendered completely normal). I'll have to unfortunately reject this one.

PR #173, opened 2020-11-12
Added effect Bouncy Vehicles

ProfessorBiddle:
Makes vehicles bounce aggressively off of collisions. May result in flight.


ChaosMod/Effects/db/Vehs/VehsBouncy.cpp
```diff
@@ -0,0 +1,27 @@
+#include <stdafx.h>
+
```
pongo1231, line 2: I think you forgot to add your credits here :)

PR #175, opened 2020-11-12
Added Effect Fake Crash

ProfessorBiddle:
Fakes a game crash with a stop, a skip forward, and another stop.

The sound is a little funky, but I can't really find a better one so if someone has a better idea let me know.


MoneyWasted:
I believe you can show the "There has been an error joining a session" message from GTA:O it would be funny to do that instead if it is possible. I will have to look into that.

ProfessorBiddle:
Implemented, but I can't get it to actually display the error text, just the big "alert" header

MoneyWasted:
In FiveM I do it like this.

`-- Make the first line using custom text.
AddTextEntry("warning_message_first_line", "This is the first line.")

-- Make the second line using custom text.
AddTextEntry("warning_message_second_line", "This is the second line!")

-- Add an event handler for when the screen is dismissed.
AddEventHandler("optionSelected", function(selected)
    print(selected) -- do whatever you want with the selected choice.
    -- players can either press the physicial buttons, or they can click
    -- the instructional buttons with their mouse and it will trigger
    -- the event as well.
end)


-- Create a thread to loop this warning message.
CreateThread(function()
    while true do
        Wait(0)
        -- Display the warning message every tick.
        SetWarningMessage("warning_message_first_line", 82, "warning_message_second_line", 0, -1, true, 0, 0, 0)
        
        -- Check for key presses or instructional button clicks.
        -- Input group of 2 is required for this to work while the warning is being displayed.
        
        if (IsControlJustReleased(2, 201) or IsControlJustReleased(2, 217)) then -- any select/confirm key was pressed.
            TriggerEvent("optionSelected", "select")
            break
        elseif (IsControlJustReleased(2, 203)) then -- spacebar/x on controller (alt option) was pressed.
            TriggerEvent("optionSelected", "alt")
            break
        elseif (IsControlJustReleased(2, 202)) then -- any of the cancel/back buttons was pressed
            TriggerEvent("optionSelected", "cancel")
            break
        end
    end
end) `

and the result is [this](https://i.imgur.com/imwoimm_d.jpg?maxwidth=640&shape=thumb&fidelity=medium)

Most likely the issue is you have to invoke the text you use. FiveM has their own called ADD_TEXT_ENTRY I don’t know if normal gta has something similar to that.

MoneyWasted:
In FiveM I do it like this.

`-- Make the first line using custom text.
AddTextEntry("warning_message_first_line", "This is the first line.")

-- Make the second line using custom text.
AddTextEntry("warning_message_second_line", "This is the second line!")

-- Draw the alert.
SetWarningMessage("warning_message_first_line", 82, "warning_message_second_line", 0, -1, true, 0, 0, 0)`

and the result is this

Most likely the issue is you have to invoke the text you use. FiveM has their own called ADD_TEXT_ENTRY I don’t know if normal gta has something similar to that.

MoneyWasted:
In FiveM I do it like this.

-- Make the first line using custom text.
AddTextEntry("warning_message_first_line", "This is the first line.")

-- Make the second line using custom text.
AddTextEntry("warning_message_second_line", "This is the second line!")

-- Draw the alert.
SetWarningMessage("warning_message_first_line", 82, "warning_message_second_line", 0, -1, true, 0, 0, 0)

and the result is [this](https://i.imgur.com/imwoimm_d.jpg?maxwidth=640&shape=thumb&fidelity=medium)

Most likely the issue is you have to invoke the text you use. FiveM has their own called ADD_TEXT_ENTRY I don’t know if normal gta has something similar to that.

drunderscore:
@MoneyWasted Sadly, your example is using custom FiveM stuff to make it work properly.
The `AddTextEntry` (from what I can tell) is creating an on-the-spot GXT entry, as `SET_WARNING_MESSAGE` only allows a GXT entry. So, we won't be able to supply our own string.

However, we still could make this work, simply without the native. The scaleform used seems to be `custom_warning_screen`, which extends from the `instructional_buttons` scaleform. So it would certainly still be possible to create it with a custom message, and even custom button, but you would just have to know how to do it.

ProfessorBiddle:
I'll look through the existing messages to see if one of those is fitting, if not the blank lines should be good enough. 

I like the button not doing anything, it makes it feel more like a hard crash

MoneyWasted:
Here seems to be some more occurrences of this native being used in GTA. https://github.com/root-cause/v-decompiled-scripts/search?q=SET_WARNING_MESSAGE

MoneyWasted:
Along with a few more being used in a similar native.
https://github.com/root-cause/v-decompiled-scripts/search?q=SET_WARNING_MESSAGE_WITH_HEADER

pongo1231:
I think a more convincing fake crash could be achieved by removing the scaleform used and instead playing the windows error sound (achieved through `PlaySound(reinterpret_cast<LPCSTR>(SND_ALIAS_SYSTEMHAND), NULL, SND_ALIAS_ID);`) and setting the game to the slowest speed possible for 3 - 5 seconds.

PR #176, opened 2020-11-12
Added effect Everyone Turn Right

ProfessorBiddle:
Makes all peds in vehicles turn right and accelerate, and locks the players steering at full right lock.  Makes traffic quite chaotic.


pongo1231:
Since it's essentially taking away player controls in vehicles I turned it into a short timed effect

PR #177, opened 2020-11-12
Added Effect Ramp Jam

ProfessorBiddle:
When you are in a vehicle and press space, it spawns a ramp in front of you and gives you a slight speed boost to make you jump. Super useful tool, until you jump straight into a wall.

Based on Ramp Jam mod https://www.gta5-mods.com/scripts/ramp-jam#:~:text=PREMISE-,Ramp%20Jam%20allows%20you%20to%20instantly%20deploy%20a%20ramp%20directly,unaccessible%20areas%20of%20the%20map.


pongo1231:
There definitely should be some kind of indication as to what the key for spawning a ramp is for players (either simply adding it to the effect name like "Vehicles Shoot Rockets" or showing a notification in-game).

pongo1231:
Actually I haven't seen it's already included in the effect name in-game, my bad.

PR #179, opened 2020-11-14
Meta: Use CMake for ASI mod

drunderscore:
CMake is a build process that is usually used to aid with platform agnostic building, but in our case allows:
- Easier merging of effects
- Faster builds (depending on build system)
- Support more IDEs

If I'm being quite honest, Visual Studio sucks. The good thing about CMake is that it still allows Visual Studio to be used, as well as other build systems to support more environments.

I certainly would like some feedback from @pongo1231 and @Last0xygen, and ofcourse anyone else who has some comments or suggestions on further improvements -- however I'm happy with this current implementation.


Last0xygen:
Looks good to me. Seems to work fine (didnt test with GTA yet), although some extra clicks are required for visual studio 😄 
This could improve those mergeconflict in project files, though it opens up a new "problem", if you don't manage the MakeList-File strictly and miss an effect.

MoneyWasted:
I'm personally a fan of the Visual Studio compiler but I am open to using CMake. Having it as a option people can use would be the best solution as I would still like to use Visual Studio's solution so I don't have to worry about adding new files to build to the CMake file.

drunderscore:
Rebased for master

ProfessorBiddle:
Unless the visual studio compiler is causing issues to the final asi I'd say just leave it, it doesn't take long to compile such a small file and setting up and using the visual studio compiler is really easy

drunderscore:
Closing this for the time being.
There is obviously not much interest, and I have not kept this up to date. If this is desired in the future, I will re-open/re-create this.

PR #180, opened 2020-11-15
Added Effect Easy Aim

ProfessorBiddle:
Slows down time while the player is aiming or shooting


ChaosMod/Effects/db/Player/PlayerEasyAim.cpp
```diff
@@ -0,0 +1,21 @@
+#include <stdafx.h>
+
+// Effect by ProfessorBiddle
+
+static void OnTick()
+{
+	static float verticalRecoil = 4;
```
Last0xygen, line 7: Copy paste fail

Last0xygen (changes requested):

pongo1231:
Should be a tad faster IMO, also I think naming it "Bullet Time" instead gets the point across even more as to what the effect actually does.

PR #183, opened 2020-11-16
Added Effect Stop and stare

ProfessorBiddle:
Makes everyone stop in place/stop their vehicles and get out, and stare at the player.


Last0xygen:
Isn’t this the same as „You are famous“?

TurnrDev:
Not really, in You are Famous, they run towards you

ProfessorBiddle:
Not really, in you are famous they continually run towards you and even
pull you out of vehicles, with this they just stop. And state. And crash
their helicopters.

ChaosMod/Effects/db/Peds/PedsStopAndStare.cpp
```diff
@@ -0,0 +1,21 @@
+#include <stdafx.h>
+static void OnStart()
```
pongo1231, line 2: Don't forget to add your credits

PR #188, opened 2020-11-25
Adjust 'Replace Vehicle'

Last0xygen:
Spawns new Random vehicle if on foot


Last0xygen:
I just realised that MP-Vehicles will respawn after 1-2 seconds. this seems to be a "known issue"
@pongo1231 do you know anything about this?

Last0xygen:
Ah yes, did mean despawn of course. Yeah, found that too, thought there may be another way 🤔 alright, will look into that :)

pongo1231:
Try setting `SET_ENTITY_AS_MISSION_ENTITY` on the vehicle. Setting it as a mission entity explicitly should fix the despawning. What could also help is setting the `netMissionEntity` flag in `CREATE_VEHICLE` to true.

pongo1231:
I made some changes which could help with the despawning, however it still happens with some vehicles. Since it's the game scripts controlling that it'll probably need changes there (might be as easy as simply setting a Global). I'll merge this since it's good enough for now.

PR #190, opened 2020-11-30
Implement Realistic Hacking

drunderscore:
Using the Hacking_PC scaleform, we emulate what would happen during The
Bureau Raid (roof entry), and insert our own custom words, win phrases,
and fail result.

A lot of moving parts to this, so I encourage a thorough review, making
sure no part can break accidentally :^)

### DIFF
--- a/ChaosMod/ChaosMod.vcxproj
+++ b/ChaosMod/ChaosMod.vcxproj
@@ -106,6 +106,7 @@
     <ClCompile Include="Effects\db\Peds\PedsMinions.cpp" />
     <ClCompile Include="Effects\db\Peds\PedsSpawnAngryJimmy.cpp" />
     <ClCompile Include="Effects\db\Player\PlayerEasyAim.cpp" />
+    <ClCompile Include="Effects\db\Player\PlayerHacking.cpp" />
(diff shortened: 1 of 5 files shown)


drunderscore:
Here's a bit of what it's like: https://streamable.com/e8dqaf

Last0xygen (changes requested):
* unnecessary EffectsInfo-Changes

Last0xygen:
Nice work!
Although i thoguht it should kill you if you didn't solve it in time, this would cause problems once this effect would be set as a short effect or with a really short time.
One thing i'm confused about: your streamable-link doesnt show any time-bar for the effect, which should be visible or am i missing something?

drunderscore:
> ...Although i thoguht it should kill you if you didn't solve it in time, this would cause problems once this effect would be set as a short effect or with a really short time.

Definitely a good idea, I'll just kill on timeout as well.

> ...One thing i'm confused about: your streamable-link doesnt show any time-bar for the effect, which should be visible or am i missing something?

It is visible, it's simply that when I loaded the mod, I paused the timer (ctrl + -), and then selected the effect from the debug menu, which reset the timer. Because the timer is paused, there isn't any progress to see :^)

pongo1231:
Good job with this one, really liking it :)

PR #202, opened 2020-12-01
Use MISC::GET_GAME_TIMER in place of GetTickCount64

drunderscore:
GetTickCount64 returns the milliseconds since system startup, which is
real-time. Although this is okay to use, in ~certain~ most cases we would
rather use game time to gain the benefits of it:
- Effected by timescale
- Stops when game is paused
- Isn't real-time :^)

Example of this being beneficial:
The effect Need for Speed is applied. Game is paused. 30 seconds later,
game is unpaused. From the perspective of GetTickCount64, 30 seconds has
passed, and so you instantly fail Need for Speed and explode. With
in-game time, however, it wasn't counting down while paused, so
unpausing will continue the timer as normal.

Another example:
The effect Need for Speed is applied, while a gamespeed effect is also
active. With GetTickCount64, you would have to accelerate to the minimum
speed in real-time 10 seconds (or less), while only having half your
acceleration speed (or less!). With in-game time, however, it's effected
by timescale, so you would be able to accelerate in the slow speed, with
the timer also being slowed.

There are a few places I didn't use this:
- Anything that ISN'T an effect
- Fake death/fake crash, as the timescale change/busy loop would make
  the wait time different than previously.

Some other effects may have been changed in a negative way as well.
Although we should rid all instances of GetTickCount64 (in effects),
let's make sure that the current changes aren't detrimental to the
effect itself.

_extra:_
If GetTickCount64 was called to initialize a static variable in global scope, I omitted it, as it would then cause the native to be executed off the native thread, which is bad!

If it was a static variable initialization _in function scope_, then it was okay, as that is only initialized once reached. _(I think...)_

### DIFF
--- a/ChaosMod/Effects/db/Misc/MiscLag.cpp
+++ b/ChaosMod/Effects/db/Misc/MiscLag.cpp
@@ -15,7 +15,7 @@ static void OnTickLag()
 {
 
 	static DWORD64 lastTick = 0;
-	DWORD64 curTick = MISC::GET_GAME_TIMER();
+	DWORD64 curTick = GET_GAME_TIMER();
 
 	if (curTick > lastTick + 500)
 	{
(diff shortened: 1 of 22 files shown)


pongo1231:
Definitely a good idea, the `MISC` namespace can be omitted since it (and all the other native namespaces) are imported in `nativesNoNamespaces.h` though.

PR #204, opened 2020-12-01
Added effect DarkViperAU that makes it rain Cola

igerlster:
Added effect DarkViperAU that makes it rain Cola, very small change from money rain effect


igerlster:
Found Errors

Issue #174, opened 2020-11-12
Catto guns almost always results in crash during shootout


ProfessorBiddle:
Catto guns almost always results in a crash, especially when in a shootout. Is there a way to remove old/dead cats after a limit is reached, or prevent cats from building up too much some other way?

pongo1231:
The cats are already flagged for despawning but seems like the game won't despawn them in time to prevent a game crash from having too many existing cats at once. Implementing a hard limit should be easy by storing them in a list, keeping track of the existing amount of cats and deleting the oldest ones accordingly.

MoneyWasted:
Add a counter for each cat shot increment by 1 and store the entity handle for that cat if there are 100 cats delete all the cats or to make it less abrupt delete the first 10 cats shot so we have a max of 100 cats spawned.

drunderscore:
This would certainly work, but it isn't the best or correct solution.
The actual problem is that we are exceeding the ped/vehicle/prop pool limits, which instantly crash the game.
Most of this information can be [read from SHVDN](https://github.com/crosire/scripthookvdotnet/pull/956).

I plan to PR a common interface for spawning entities, that will ensure the pool limit it never exceeded. SHVDN is licensed under zlib, so as long as we credit them (and kagikn in this case), we should be good.

PR #214, opened 2020-12-04
Attempt at fixing crash with 'Catto Guns'

ProfessorBiddle:
Fix: #174

Ater the first 10 cats it starts deleting objects that aren't marked as mission entities, this prevented crashes in all of my testing and didn't effect missions much except making traffic/pedestrians a little less common.

I don't want to delete just cats because that would minimize the impact of the effect.

This isn't perfect, and may impact other effects such as whale rain/spawning other stuff, but it seems to work fairly well.

### DIFF
--- a/ChaosMod/Effects/db/Peds/PedsCatGuns.cpp
+++ b/ChaosMod/Effects/db/Peds/PedsCatGuns.cpp
@@ -44,6 +44,7 @@ static void OnTick()
 {
 	static const Hash catHash = GET_HASH_KEY("a_c_cat_01");
 	LoadModel(catHash);
+	int numCats = 0;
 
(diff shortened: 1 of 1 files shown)


Last0xygen:
It should really just remove other cat entities, not ANY loaded peds

ProfessorBiddle:
> It should really just remove other cat entities, not ANY loaded peds

I tested in multiple missions (mr phillips, blitz play, paleto score) and it doesn't remove peds that are involved in the mission, mostly peds outside of the mission and cats. I didn't want to remove only cats because having a couple hundred cats is hilarious.

pongo1231:
Increasing the limit of cats to around 20 and simply removing cats which spawned first should work just fine, no need to remove other existing peds.

Issue #181, opened 2020-11-16
Exploit in Simeon Says.


LennartOschika:
When there is Simeone Says and the "task" is "move right" for example it is possible to hold both the left and right button. 

This results in one being able to move straight forward which kind of defeats the point of the effect. Of course you still can't move left but if you also couldn't use this exploit it is harder.

PR #215, opened 2020-12-04
Remove exploit in Simeon Says

ProfessorBiddle:
fix: #181


Last0xygen:
this does not remove the exploit itself, but explode you if you press either of the opposite direction-keys. i don't think this is the solution we should be using

ProfessorBiddle:
"Move Left" does pretty much mean not moving right, this is the best way I
could think of to enforce always moving the correct direction.

If you have a better suggestion I'd love to redo it, this solution
definetely isn't optimal.

Issue #205, opened 2020-12-01
Bus Bois crashing game


DestroyerXyz:
Bus bois crashing the Steam version of the game.

ProfessorBiddle:
I don't have the steam version, so if someone else could test this that would be fantastic, it works fine on the Epic Games version where I made it.

DestroyerXyz:
DarkViperAU crashes 2 times in this video because of Bus Bois: https://www.youtube.com/watch?v=o8OScme2T3E&ab_channel=DarkViperAU

DestroyerXyz:
I think he has a Threadripper 3970X and an RTX 2080Ti

ProfessorBiddle:
I meant if someone could test for a solution, it should be fairly easy to track down the issue if someone has the right game version.

DestroyerXyz:
I think he uses the latest game version for Chaos

DestroyerXyz:
I myself have the Epic Games Version😅

MoneyWasted:
It seems that we are hitting some limit internally. Most likely a polygon limit due to the amount of vehicles.

It could also be that creating a bunch of explosions in one area crashes people. This is a common method used as a "crash exploit" in GTA Online mod menus.

DestroyerXyz:
What does Bus Bois actually do? I could not see what it did because it crashed.

OHSHIET:
#167: puts peds in buses.

pongo1231:
Haven't had any issues with it on the Steam version. I've tested it multiple times on the highway with traffic density set to highest, which should be the absolute worst case. I've also seen several streams / videos where that effect works without any issues, which makes it even more confusing as to why it caused a game crash for him every time. I wonder if it's some mod conflicting with that effect (which would be very unlikely unless it's either modifying every existing vehicle in some way or hooking `CREATE_VEHICLE` for some reason).

PR #217, opened 2020-12-05
Attempt at fixing bus bois lag/crash

ProfessorBiddle:
Fix: #205

Checks for distance from player now, should reduce the number of entities in buses and hopefully prevent crashes.

Not sure what to set the variable at as I've never had lag/crashing issues, but if someone who is more familiar with GTA coordinate units and how they relate to the actual map has a better suggestion let me know.


DestroyerXyz:
CPP is hard AF

Issue #212, opened 2020-12-03
[Suggestion] Effect: Spawn 10 Buffalos


JustTunerYT:
So, i remembered the glitch in race event, that Matt (aka DarkViperAU) loved, which spawns buffalo every time player repeats race after fail (https://www.youtube.com/watch?v=RpmkZ2d1tbw). I thought of it and decided to suggest spawning these 10 times. Color: White. Upgrades: None

TheRealAyCe:
Alternative names:
Buffalo Stampede
Car Herder

ProfessorBiddle:
Alternately spawning a buffalo every time the player interacts with something (using phone, entering vehicle, starting mission, reaching checkpoint) would be interesting,

JustTunerYT:
I think negative, because if it will affect shooting too, game would crash

ProfessorBiddle:
Good point, maybe have a timed effect that spawns a buffalo every few
seconds?

JustTunerYT:
But, what about this one? I thought, that it would be exactly 10 buffalos and it will be one timed effect

JustTunerYT:
Oh, nvm. Had to press "more" button.
Closed

PR #218, opened 2020-12-05
Added Effect GTA Racing

ProfessorBiddle:
Fix: #212

Spawns a white buffalo every 5 seconds, and sets the player into it.

### DIFF
--- a/ChaosMod/Effects/db/Vehs/VehsGtaRacing.cpp
+++ b/ChaosMod/Effects/db/Vehs/VehsGtaRacing.cpp
@@ -1,0 +1,22 @@
+#include <stdafx.h>
+
+//Effect by ProfessorBiddle, suggested by JustTunerYT
+
+static void OnTick()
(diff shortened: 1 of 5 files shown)


ChaosMod/Effects/db/Vehs/VehsGtaRacing.cpp
```diff
@@ -0,0 +1,22 @@
+#include <stdafx.h>
+
+//Effect by ProfessorBiddle, suggested by JustTunerYT
+
+static void OnTick()
+{
+	static auto lastTick = GetTickCount64();
+	auto curTick = GetTickCount64();
```
Last0xygen, line 8: You should switch to "GET_GAME_TIMER". For more info on the differences see https://github.com/gta-chaos-mod/ChaosModV/pull/202

Last0xygen (changes requested):

PR #224, opened 2020-12-07
Added Effect 'Quake FOV'

Last0xygen:


MoneyWasted:
I think “Quake Pro” would be a suitable name. It’s the name given to the highest FOV in Minecraft. (120 degrees and above.)

PR #228, opened 2020-12-07
Added Effect 'Diving Intervention'

ProfessorBiddle:
Finally an extremely helpful effect!

Spawns 20 very strong Jesus' with powerful weapons to defend you.

This is not programmed super efficiently, but it works really well and doesn't cause any significant lag. If you have advice on trimming it down I welcome it

### DIFF
--- a/ChaosMod/Effects/db/Peds/PedsIntervention.cpp
+++ b/ChaosMod/Effects/db/Peds/PedsIntervention.cpp
@@ -1,0 +1,87 @@
+#include <stdafx.h>
+
+//Effect by ProfessorBiddle
+
+
+//future plans: make jesus spawn above and parachute to player, possibly jesus sniper in helicopter
+static void OnStart()
+{
+	Ped playerPed = PLAYER_PED_ID();
+	Vector3 playerPos = GET_ENTITY_COORDS(playerPed, true);
+	static constexpr Hash modelHash = -835930287;
+
+	Hash relationshipGroup;
+	ADD_RELATIONSHIP_GROUP("_DIVINE_JESUS", &relationshipGroup);
+	SET_RELATIONSHIP_BETWEEN_GROUPS(0, relationshipGroup, GET_HASH_KEY("PLAYER"));
+	SET_RELATIONSHIP_BETWEEN_GROUPS(0, GET_HASH_KEY("PLAYER"), relationshipGroup);
+	for (int i = 0; i < 4; i++)
+	{
+		//I am shit at programming if someone wants to do this in a more efficient way please do
+		//spawn 5 jesus
+		Ped ped0 = CreatePoolPed(4, modelHash, playerPos.x, playerPos.y, playerPos.z, GET_ENTITY_HEADING(playerPed));
+		Ped ped1 = CreatePoolPed(4, modelHash, playerPos.x + 5.f, playerPos.y, playerPos.z, GET_ENTITY_HEADING(playerPed));
+		Ped ped2 = CreatePoolPed(4, modelHash, playerPos.x - 5.f, playerPos.y, playerPos.z, GET_ENTITY_HEADING(playerPed));
+		Ped ped3 = CreatePoolPed(4, modelHash, playerPos.x, playerPos.y, playerPos.z + 5.f, GET_ENTITY_HEADING(playerPed));
+		Ped ped4 = CreatePoolPed(4, modelHash, playerPos.x, playerPos.y, playerPos.z - 5.f, GET_ENTITY_HEADING(playerPed));
+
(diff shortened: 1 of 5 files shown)


ChaosMod/Effects/db/Peds/PedsIntervention.cpp
```diff
@@ -0,0 +1,87 @@
...
+		SET_PED_COMBAT_ATTRIBUTES(ped0, 5, true);
+		SET_PED_COMBAT_ATTRIBUTES(ped0, 46, true);
+		SET_PED_ACCURACY(ped0, 100);
+		SET_PED_FIRING_PATTERN(ped0, 0xC6EE6B4C);
+
+		SET_PED_SUFFERS_CRITICAL_HITS(ped1, false);
+		SET_PED_RELATIONSHIP_GROUP_HASH(ped1, relationshipGroup);
+		SET_PED_HEARING_RANGE(ped1, 9999.f);
+		SET_PED_AS_GROUP_MEMBER(ped1, playerGroup);
+		SET_ENTITY_PROOFS(ped0, true, true, true, true, true, true, 1, true);
```
kolyaventuri, line 43: Should this not be `ped1`? Or am I mistaken about what `SET_ENTITY_PROOFS` is doing here? Just diving into the code + PRs for this project 🙂 
 
```suggestion
		SET_ENTITY_PROOFS(ped1, true, true, true, true, true, true, 1, true);
```

Also L53, L63, L73

ChaosMod/Effects/db/Peds/PedsIntervention.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
+	Ped playerPed = PLAYER_PED_ID();
+	Vector3 playerPos = GET_ENTITY_COORDS(playerPed, true);
+	static constexpr Hash modelHash = -835930287;
+
+	Hash relationshipGroup;
+	ADD_RELATIONSHIP_GROUP("_DIVINE_JESUS", &relationshipGroup);
+	SET_RELATIONSHIP_BETWEEN_GROUPS(0, relationshipGroup, GET_HASH_KEY("PLAYER"));
+	SET_RELATIONSHIP_BETWEEN_GROUPS(0, GET_HASH_KEY("PLAYER"), relationshipGroup);
+	for (int i = 0; i < 4; i++)
```
kolyaventuri, line 17: Could this not be accomplished as a single loop of 20, _or_ with a nested for loop (outer being 4, inner being 5)? Given that a lot of things are repetitive, feels like you could just be calling `SET_PED_FIRING_PATTERN` based off a Hash array.

Something like...
```cpp
Hash patterns[] = {0xC6EE6B4C, ...};

for (int i = 0; i < 20; i++) {
  Ped ped = CreatePoolPed(...);
  
  SET_PED_SUFFERS_CRITICAL_HITS(ped, false);
  ...
  SET_PED_FIRING_PATTERN(ped, pattern[i % 5]);
}
```

I'm not up to date on my C++, but. Thinking out loud.

ProfessorBiddle: I was trying to do sets of 5 each with different weapons and firing patterns which would be hard in a loop, but it could probably be done in a loop of 20 yeah

ChaosMod/Effects/db/Peds/PedsIntervention.cpp
ProfessorBiddle: Yes it should be, I will fix that

MoneyWasted:
You can make all the peds parachute to a given location using TASK_PARACHUTE_TO_TARGET.

Spawn the peds above the player and specify the player cords in TASK_PARACHUTE_TO_TARGET like this...

TASK_PARACHUTE_TO_TARGET(ped, player.x, player.y, player.z);

ChaosMod/Effects/db/Peds/PedsIntervention.cpp
kolyaventuri: If you nested the loops though, could you not achieve the same thing in less code?

```cpp
/* Group */
for (int i = 0; i < 4; i++) {
  /* Peds */
  for (int j = 0; j < 5; j++) { 
    Ped ped = CreatePoolPed(...);
    ...
  }
}
```

Since at least right now, you have 4 identical groups, no? So 5 peds with `WEAPON_RAYPISTOL` and firing pattern `0xC6EE6B4C` and so on and so forth

ProfessorBiddle:
**> You can make all the peds parachute to a given location using TASK_PARACHUTE_TO_TARGET.

I did this, but they kept closing after a few seconds and falling to the ground, I think I need to set the relationship groups after they touch the ground

Issue #216, opened 2020-12-05
Suggestion: Add AI blips to spawned entities


ProfessorBiddle:
Only to spawned enemies/companions, not vehicles/whales etc. It isn't hard to implement, just looking for you guys' thoughts before I do.

MoneyWasted:
I'll get this implemented.

pongo1231:
IMO the uncertainty of the spawned enemies' whereabouts adds to the fun. Very unlikely to be implemented.

PR #229, opened 2020-12-08
Adding Blips to Spawned Peds

MoneyWasted:
Pretty self-explanatory. It only adds blips to peds that don’t have one by default. Hostile peds have a red blip and companions have green blips.


MoneyWasted:
https://github.com/gta-chaos-mod/ChaosModV/issues/216

PR #233, opened 2020-12-08
Adding Blips to Spawned Peds

MoneyWasted:
I messed up my last repo by committing changes to the master branch (stupid mistake I know). I fixed my mistakes and committed the changes to a different branch.


TheRealAyCe:
Wouldn't it be better to create a function for that to be called in each of those cpps instead of including 3 lines every time?

MoneyWasted:
> Wouldn't it be better to create a function for that to be called in each of those cpps instead of including 3 lines every time?

Yea I could do that but it wouldn't be that hard for someone to add three lines in and I don't think there will be to many new effects that spawn peds.

ProfessorBiddle (approved):

pongo1231:
I think part of the fun of the enemies is not knowing where exactly they are located which makes them harder to deal with. Also it won't really work with allies since those will stop acting as companions after a certain distance and can even turn into enemies at certain times (e.g. when "All Peds Attack Player" is active). I don't see this as suitable unfortunately.

PR #289, opened 2020-12-29
Add Effect: Bees

### LINKED ISSUE #237: [Suggestion] Effect: Bees
- Little dots (black?) "fly" around the player (not sure of the implementation details on this, but I imagine somewhat similar to [MiscDVDScreensaver.cpp](https://github.com/gta-chaos-mod/ChaosModV/blob/master/ChaosMod/Effects/db/Misc/MiscDVDScreensaver.cpp), combined with the active players screen X / Y coords)
- Player takes 1hp (?) of damage every now and then as they "sting" you

I'd be happy to take this on at some point, just thought I'd throw out the idea before diving in. Not sure when I can get to it though.

kolyaventuri:
fixes #237

Adds effect `Bees`

- Spawns a cloud of "bees" (`ent_amb_fly_swarm`)
- Randomly damages the player by 1 hp every now and then, as if the bees are "stinging" the player

**TODO**
- Add a damage sound when the bees sting the player. For whatever reason, neither `SET_ENTITY_HEALTH` nor `APPLY_DAMAGE_TO_PED` achieve this


ChaosMod/Effects/EffectsInfo.h
```diff
@@ -512,6 +513,8 @@ const std::unordered_map<EffectType, EffectInfo> g_effectsMap =
 	{EFFECT_PEDS_FLIP, {"Spinning Peds", "peds_flip", true}},
 	{EFFECT_PLAYER_PACIFIST, {"Pacifist", "player_pacifist", true, {}, false}},
 	{EFFECT_VEH_TURN_RIGHT, {"Everyone Turn Right", "veh_turnright", true, {}, true}},
+	{EFFECT_PEDS_BUSBOIS, {"Bus Bois", "peds_busbois"}},
     {EFFECT_PEDS_BUSBOIS, {"Bus Bois", "peds_busbois"}},
     {EFFECT_PLAYER_HACKING, {"Realistic Hacking", "player_hacking"}},
```
kolyaventuri, line 518: Not sure what happened with the whitespace here. Will resolve

ChaosMod/Effects/db/Player/PlayerBees.cpp
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// (kolyaventuri) There's gotta be a better way of doing this...
+static void OnTick() {
+    Ped player = PLAYER_PED_ID();
+    int rand_int = g_random.GetRandomInt(0, CHANCE);
+
+    if (rand_int == match) {
+        APPLY_DAMAGE_TO_PED(player, 1, false, false);
+        // TOOD: Add a damage sound here
```
kolyaventuri, line 29: If anyone's aware of a native I can use here to trigger the damage sound, or the IDs of those sounds, that would be great. Or a better native to use :) 

Thinking, would want Michael / Trevor / Franklin's respective damage sounds. The groans / etc. that you hear when they get hit

MoneyWasted:
Looks good having the pain sound effects is a nice touch. Haven’t look into it but it’s definitely possible, if the player is “stung” they should get some sort of short visual effect like a red flash around the screen implying pain and maybe a slight shake of the screen.

kolyaventuri:
@MoneyWasted oh that'd be fun. I'll dig into it a bit and see if it's doable

MoneyWasted:
Awesome! The vignette adds a nice touch especially with the use of the time cycle modifier. Clever thinking. 👍🏻

PR #305, opened 2021-01-02
Adjusted EntityHandleHook, ConfigApp and ReplaceVehicle

Last0xygen:
* Config-App didn't allow removing a set seed
* "Random Seed" wasn't really 'random'
* "Replace Vehicle" often didn't work as inteded ("Mission Failed - 'VEHICLE' was destroyed")
  * Vehicle deletion was too far apart from the hook-replacement
  * Now Keeping all Entity-Replacements in a maptable
  * This allows several replacements to be run
  * Performance should be checked again, i only noticed a dip in FPS after ~ 15/20 Entries for the SAME vehicle. After that entity was removed, got all the FPS back
* Removed unnecessary Vehicle struct in effect 'Beyblade'


ChaosMod/Main.cpp
```diff
@@ -25,7 +25,7 @@ void ParseConfigFile(int& effectSpawnTime, int& effectTimedDur, int& seed, int&
 
 	effectSpawnTime = configFile.ReadValueInt("NewEffectSpawnTime", 30);
 	effectTimedDur = configFile.ReadValueInt("EffectTimedDur", 90);
-	seed = configFile.ReadValueInt("Seed", 0);
+	seed = configFile.ReadValueInt("Seed", g_random.GetRandomInt(0, MAXINT));
```
pongo1231, line 28: This will cause the mod to always use this seed generated by the config utility. It uses a random seed internally if the seed is set to 0 anyways (see Random.h).

Last0xygen: No, this is inside the ChaosMod, not the config-app. But yeah, if its already handled in random.h it should not be needed anymore 😅

Last0xygen: Fixed :)

ChaosMod/Memory/Hooks/HandleToEntityStructHook.cpp
```diff
@@ -1,7 +1,6 @@
 #include <stdafx.h>
 
-static Entity m_origHandle = 0;
-static Entity m_proxyHandle = 0;
+static std::map<Entity, Entity> vehicleMap;
```
pongo1231, line 3: A hash map (`unordered_map`) would be more fitting here. Also it'll need to be cleaned up regularly by removing all deleted entities from it to avoid it from growing infinitely.

ChaosMod/Memory/Hooks/HandleToEntityStructHook.cpp
```diff
@@ -10,13 +9,16 @@ __int64 _HK_HandleToEntityStruct(Entity entity)
 	{
 		return 0;
 	}
+	Entity vehToContinue = entity;
 
-	if (entity == m_origHandle)
+	while (vehicleMap.count(vehToContinue) > 0)
```
pongo1231, line 14: Shouldn't this be an `if` instead of `while`? (Also an `empty` function exists already which does the same check)

ConfigApp/MainWindow.xaml.cs
```diff
@@ -411,6 +411,10 @@ private void user_save_Click(object sender, RoutedEventArgs e)
             WriteTwitchFile();
             WriteEffectsFile();
 
+            // Reload saved config to show the "new" (saved) settings
+            ParseConfigFile();
+            ParseTwitchFile();
```
pongo1231, line 416: This shouldn't be necessary since it fetches all the user inputs to write the config files and doesn't touch them in any way (so the user inputs should correspond to what's in the config files anyways).

pongo1231 (changes requested):

ChaosMod/Memory/Hooks/HandleToEntityStructHook.cpp
Last0xygen: nah, im resolving the id here, as it can be multiple layers.
id 1 orig -> id 2
id 2 -> id 3
id 3 -> id 4

asking for id 1 would therefore be resolved to id4 :)

pongo1231: Ah I see. Also disregard my latter statement, I didn't see the `vehToContinue` passed to `count`.

ConfigApp/MainWindow.xaml.cs
Last0xygen: it DOES change it though, if you empty a textbox (timer duration for example) and saving it. thats what i noticed :)

pongo1231: Yeah it doesn't update the text boxes if an user enters invalid input and saves it currently. Makes sense to read the config files again to update the views in those cases.

Last0xygen:
Fixed everything, but now the EntityHook is working TOO good 😂 
The problem is: after applying the hook, the old vehicle should be deleted. This results in getting resolved to the new vehicle and deleting thjat instead of the old one.
i currently don't know how to tackle this :|

pongo1231:
Simply delete the vehicle before applying the hook, it won't cause any mission fails since the mission scripts don't have a chance to run between those two calls (since there's no `WAIT` call).

ChaosMod/Memory/Hooks/HandleToEntityStructHook.cpp
```diff
@@ -10,13 +9,15 @@ __int64 _HK_HandleToEntityStruct(Entity entity)
...
 		return 0;
 	}
-
-	if (entity == m_origHandle)
+	Entity vehToContinue = entity;
+	while (vehicleMap.count(vehToContinue) > 0)
 	{
-		return _OG_HandleToEntityStruct(m_proxyHandle);
+		vehToContinue = vehicleMap[vehToContinue];
+    	if (vehToContinue <= 0) {
```
pongo1231, line 16: The bracket here should be on a new line

Last0xygen:
> Simply delete the vehicle before applying the hook, it won't cause any mission fails since the mission scripts don't have a chance to run between those two calls (since there's no `WAIT` call).

Still fails the mission for me when I do that

ChaosMod/Memory/Hooks/HandleToEntityStructHook.cpp
pongo1231: That's fixed but now its indentation is broken 😅

Last0xygen: oh the joy of editing via mobile 👌

Issue #236, opened 2020-12-09
[Suggestion] Feature: Metaeffects


SeanMirrsen:
It occurred to me that Chaos isn't quite chaotic enough.

As an optional feature, there could be a new class of effect called Metaeffects.

These effects don't mod the game, but rather mod the mod - hence, 'meta'.
By default, it could be set so that once a minute or two, there is a 50% chance of a metaeffect activating, with a standard duration of 30 seconds to 1 minute. The idea is to add metaeffects periodically, but not bombard the player with them.

Metaeffects would include:
- Doubletime/Halftime: For the duration of the effect, the timer runs at double or half speed.
- Combo time: Two effects are picked at next timer threshold instead of one.
- More the merrier: If an effect spawns an object, vehicle, or character, it spawns them twice.
- Calm before the storm: Skip adding effects for next two timers, then add three effects at once on the third.
- I'm sorry what: Hide the timer and active effects display (basically all of the mod's HUD besides the metaeffect name) for the duration.

Among possibly others I can't quite imagine at the moment.

OHSHIET:
That's a really good idea (first mistyped goood, then god, guess this represents it well enough), I was thinking the same (Chaos isn't chaotic enough), and had the same idea about changing the duration, but I just knew it wouldn't fit into the mod's atmosphere with the current effects. However I still think it needs to be implemented one way or another, because right now it's just a massive barrier which restricts us from adding more really interesting stuff.

First idea that came to mind was another bar, but a slow one. This bar would be responsible for metaeffects. For Twitch voting there would be different numbers for choosing metaeffects. however the problem is that there might be too much stuff on the screen and a lot of terminology for people who just want to have fun. Also a lot of metaeffects will be needed for a separate bar (and I guess not only for it, just for more variety...)

SeanMirrsen:
I think that, for several reasons, the metaeffects should not be on their own (visible) timer, and should not be votable.

Firstly, the entire point of the idea is to introduce less predictable variety into the mod. Metaeffects should not activate at predictable intervals, which is why I think there should be a 50% chance of one activating at the cutoff. It both gives the player breathing room, and adds an extra degree of uncertainty.
(which would be a nice positive metaeffect I think, "breathing room", pausing the timer for one cycle)

As an alternative to a second timer, perhaps metaeffects could have a lower chance of activating (20%-ish) on every cycle of the regular timer, as long as there isn't another active metaeffect.

[...]

PR #307, opened 2021-01-02
Added Metaeffects

Last0xygen:
* Added a new Category "Meta"
* Extended the Config App to modify the delay between meta-effect.
  * Didn't put too much time into it, probably should be a separate tab tbh
  * Also the effect configuration has to be adjusted, as "block from twitch voting" and "weight" doesnt make sense for them
* Added some effects
  * Change Timer-Speed (x0.5, x2, x5)
  * Change effect duration (x0.5, x2)
  * "What's Happening??" (Hide Chaos UI)
  * "Combo Time" (Adds 1-2 Random effects every time a new effect gets added)

If it's good, ill extend the config-app with another tab
Fixes https://github.com/gta-chaos-mod/ChaosModV/issues/236


ChaosMod/EffectDispatcher.cpp
```diff
@@ -86,10 +86,18 @@ void EffectDispatcher::UpdateTimer()
 		delta = 0;
 	}
 
-	if ((m_percentage = (delta + (m_timerTimerRuns * 1000)) / (m_effectSpawnTime * 1000)) > 1.f && m_dispatchEffectsOnTimer)
+	if ((m_percentage = (delta + (m_timerTimerRuns * 1000)) / (m_effectSpawnTime / g_metaInfo.timerSpeedModifier * 1000)) > 1.f && m_dispatchEffectsOnTimer)
```
kolyaventuri, line 89: I would imagine a modifier should be a multiplier, rather than something to divide by, no? Represent 2x speed by a 50% duration modifier (i.e, `time * 0.5f`)?

Last0xygen: I don’t really get what you‘re saying. 
If it would be „multiplier“ then I would agree it could be assumed to be multiplied, but as a modifier...? 
i could make it (1/modifier), which would be the same 😅

SeanMirrsen:
One thing I don't quite understand is why the 15fps/30fps limits added with the latest commit are considered "meta". The concept for metaeffects is altering the way the mod itself works, which the FPS limit doesn't seem to do.

I don't mind them being there, but they seem like just another effect.

pongo1231:
I agree with the above, the FPS related effects should be categorized as normal ones instead of meta effects.

ChaosMod/Effects/Effect.h
```diff
@@ -126,4 +126,5 @@ struct EffectData
 	bool Permanent;
 	bool ExcludedFromVoting;
 	std::string Name;
+	bool isMeta;
```
pongo1231, line 129: Should be uppercase like the other members for consistency

ChaosMod/Effects/EffectsInfo.h
```diff
@@ -254,21 +254,37 @@ enum EffectType
...
+	EffectInfo(const char* name, const char* id, bool isTimed = false, std::vector<EffectType> incompatibleList = {}, bool shortDur = false, EffectExecutionType type = EffectExecutionType::DEFAULT)
+		: Name(name), Id(id), IsTimed(isTimed), IsShortDuration(shortDur), IncompatibleWith(incompatibleList), executionType(type) {}
 
 public:
 	const char* Name;
 	const char* Id;
 	const bool IsTimed;
 	const bool IsShortDuration;
 	const std::vector<EffectType> IncompatibleWith;
+	const EffectExecutionType executionType;
```
pongo1231, line 287: This one should also be uppercase

ChaosMod/Main.cpp
```diff
@@ -163,6 +165,7 @@ void Main::Init()
 	bool enableTwitchPollVoting = stat("chaosmod/.twitchpoll", &temp) != -1 && false; // disable polls for now
 	m_twitchVoting = std::make_unique<TwitchVoting>(enableTwitchVoting, twitchSecsBeforeChatVoting, enableTwitchPollVoting, twitchOverlayMode, enableTwitchChanceSystem,
 		enableVotingChanceSystemRetainChance, enableTwitchRandomEffectVoteable);
+	g_metaInfo = MetaEffectInfo();
```
pongo1231, line 168: This line isn't needed

ChaosMod/Main.h
```diff
@@ -3,12 +3,14 @@
...
 #include "EffectDispatcher.h"
 #include "TwitchVoting.h"
+#include "Effects/MetaEffectInfo.h"
 
 #include <Windows.h>
 #include <memory>
 #include <map>
 
 inline std::map<EffectType, EffectData> g_enabledEffects;
+inline MetaEffectInfo g_metaInfo = MetaEffectInfo();
```
pongo1231, line 13: No need to set it specifically, it'll initialize as a `MetaEffectInfo` object and run the default constructor anyways.

pongo1231 (changes requested):

ChaosMod/Main.h
TheRealAyCe: Thanks, that's a pet peeve of mine, especially when "style guidelines" specifically recommend the extra line version. :D

Last0xygen:
Alright, this should be it :)

pongo1231 (approved):

PR #308, opened 2021-01-03
Add Effect: Busted

kolyaventuri:
- Forces the player to be busted
- If the player is in a car, the car is stopped
- If the player is in a non-car vehicle (plane, heli, boat, etc.), TP them to a police station then bust them
  - Hook into the fake TP methods, so the TP alone doesn't kill a mission
  - This isn't the ideal solution, but it'll work

### DIFF
--- a/ChaosMod/Effects/db/Player/PlayerBusted.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerBusted.cpp
@@ -1,4 +1,9 @@
 #include <stdafx.h>
+#include "Memory/Hooks/EntityCoordsHook.h"
+
+/*
+* Effect by kolyaventuri
+*/
 
 enum ArrestState {
 	start = 0,
@@ -16,6 +21,7 @@ static int nextModeTime = 0;
 static void OnStart() {
 	Player playerId = PLAYER_ID();
 	Ped player = PLAYER_PED_ID();
+	Vector3 pos = GET_ENTITY_COORDS(player, true);
 	const char* hashKey = cops[g_random.GetRandomInt(0, 2)];
 
 	currentMode = ArrestState::start;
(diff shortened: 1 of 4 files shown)


MoneyWasted:
You could use the same concept that “Walk on Water” uses. Spawn a solid invisible prop so entity’s and such can walk on a surface. Could also be done in the air just calculate the vehicle height and subtract from the center of the vehicle to get the z axis position where the prop should be. https://github.com/gta-chaos-mod/ChaosModV/blob/master/ChaosMod/Effects/db/Player/PlayerWalkOnWater.cpp

ChaosMod/Effects/db/Player/PlayerBusted.cpp
```diff
@@ -0,0 +1,156 @@
...
+	lastModeTime = 0;
+	nextModeTime = 0;
+
+	Ped cop = CreatePoolPed(6, GET_HASH_KEY(hashKey), 0.f, 0.f, 0.f, GET_ENTITY_HEADING(player));
+	SET_PED_AS_COP(cop, true);
+	GIVE_WEAPON_TO_PED(cop, GET_HASH_KEY("WEAPON_PISTOL"), 100, false, true); // (kolyaventuri): Cops must have weapon wielded for the arrest task to fire
+
+	Vector3 minSize;
+	Vector3 maxSize;
+	Vehicle veh;
```
pongo1231, line 37: Make sure to initialize this properly by setting it to 0

ChaosMod/Effects/db/Player/PlayerBusted.cpp
```diff
@@ -0,0 +1,156 @@
...
+	Player playerId = PLAYER_ID();
+	Ped player = PLAYER_PED_ID();
+	Vector3 pos = GET_ENTITY_COORDS(player, true);
+	const char* hashKey = cops[g_random.GetRandomInt(0, 2)];
+
+	currentMode = ArrestState::start;
+	lastModeTime = 0;
+	nextModeTime = 0;
+
+	Ped cop = CreatePoolPed(6, GET_HASH_KEY(hashKey), 0.f, 0.f, 0.f, GET_ENTITY_HEADING(player));
```
pongo1231, line 31: Doesn't have to be in the mod's entity pool since the ped is being removed manually by the effect afterwards, can simply use `CREATE_ENTITY` for that (and make sure to load the model first with `LoadModel`)

pongo1231 (changes requested):

pongo1231:
Also doesn't seem to be working very well with aircraft currently, it'll just force you out and let you fall to your death. Could also spawn an invisible platform below the player in that case.

PR #319, opened 2021-01-06
[Effect:VR] Ensure camera effect is maintained despite death

### LINKED ISSUE #318: [Bug] Virtual Reality screen effect does not persist after death
**Describe the bug**
After dying during virtual reality, the screen scanlines effect does not persist

**To Reproduce**
- Enable `Virtual Reality`
- Die

kolyaventuri:
fixed #318 

- Ensures that during VR effect, the camera scanline effect is maintained through death / respawn


kolyaventuri:
The diff on this looks kinda weird all things considered. In reality it's just making sure that the `SET_TRANSITION_TIMECYCLE_MODIFIER` call is happening during `OnTick` rather than just `OnStart`

PR #325, opened 2021-01-07
[Improvement] Make fake effects (death / teleport / player swap) have fake names

kolyaventuri:
- Adds ability for effects to have "Fake names"
- Ensures that something like "Fake Death" shows "Suicide" while the effect is running, improving the illusion
- Enables fake teleport to show the "real" teleport name briefly, and switch back to "Fake Teleport" when the fake is over

Demos:
Fake Death -> https://streamable.com/7smm80
Fake Teleport -> https://streamable.com/npxkf2

**Note**
This change is to be scoped purely to the **names** of the effects. Other changes such as fake wanted levels, fake mission failures, etc. are out of scope, and should be treated as separate adjustments since there's considerations to take in each of those (missions where you can't get the cops, whether or not a teleport should fail you, etc).


ChaosMod/Effects/EffectsInfo.h
```diff
@@ -506,13 +507,13 @@ const std::unordered_map<EffectType, EffectInfo> g_effectsMap =
 	{EFFECT_PLAYER_KEEP_RUNNING, {"Help My W Key Is Stuck", "player_keeprunning", true}},
 	{EFFECT_VEH_WEAPONS, {"Vehicles Shoot Rockets (Left Click/RB)", "veh_weapons", true}},
 	{EFFECT_MISC_AIRSTRIKE, {"Airstrike Inbound", "misc_airstrike", true}},
-	{EFFECT_FAKE_PLAYER_SWAP, {"Player Swap", "player_playerswap", true, {}, true}},
+	{EFFECT_FAKE_PLAYER_SWAP, {"Player Swap", "player_playerswap", true, {}, true, EffectExecutionType::DEFAULT, "Player Swap"}},
```
kolyaventuri, line 510: This seems redundant, however it's the most straight-forward way of the 2 (in my opinion, over `EffectDispatcher::OverrideEffectName`) to get the effect name to show up immediately

SeanMirrsen:
Maybe you could add a fix/improvement for "Fake Crash" as well? A lot of the time when the mod crashes, it has just enough time to display the name of the effect that crashed it.

It could pick from a random list of names, including "Bus Bois", "Teleport Everything to Player", "Doomsday", or some others.

kolyaventuri:
> Maybe you could add a fix/improvement for "Fake Crash" as well? A lot of the time when the mod crashes, it has just enough time to display the name of the effect that crashed it.
> 
> It could pick from a random list of names, including "Bus Bois", "Teleport Everything to Player", "Doomsday", or some others.

I've never run into this myself, but this sounds slightly out of scope for what we're trying to do here. That sounds more like a "Fake Mod Crash" effect, which would be distinctly different.

TheRealAyCe:
I like the idea though. If activating an effect crashes the game, it would be very helpful to make sure the name is displayed slightly before the effect is activated. Of course not for fake effects (other than for misdirection), but it could ease debugging.

kolyaventuri:
> I like the idea though. If activating an effect crashes the game, it would be very helpful to make sure the name is displayed slightly before the effect is activated. Of course not for fake effects (other than for misdirection), but it could ease debugging.

Effects that cause a game or mod crash should already be logged at the bottom of the `chaosmod/effectslog.txt` file

SeanMirrsen:
> 
> 
> > Maybe you could add a fix/improvement for "Fake Crash" as well? A lot of the time when the mod crashes, it has just enough time to display the name of the effect that crashed it.
> > It could pick from a random list of names, including "Bus Bois", "Teleport Everything to Player", "Doomsday", or some others.
> 
> I've never run into this myself, but this sounds slightly out of scope for what we're trying to do here. That sounds more like a "Fake Mod Crash" effect, which would be distinctly different.

I just noticed that, with the new sounds, observant streamers like DarkViperAU notice the fakery because an effect sound plays, but no effect name appears when the game "crashes". Back when Bus Bois was causing the game to crash, the effect's name appeared on the screen as it did so. This change would emulate that effect.

OHSHIET:
I feel like fake teleport should also consider the location where you got teleported, for example, give you the cops if it was Fort or the Airport, or don't give you if you're on a mission on which you can't get the cops

SeanMirrsen:
> 
> 
> I feel like fake teleport should also consider the location where you got teleported, for example, give you the cops if it was Fort or the Airport, or don't give you if you're on a mission on which you can't get the cops

I'm pretty sure that unless you can give a "fake" wanted level, giving a wanted level during a mission where a wanted level fails you, will fail the mission even if you take the wanted level back afterwards. Thus defeating the 'fakeness' of the fake teleport.

If you can somehow hook into the mission state and prevent the mission from failing until the effects are reverted, then it might work.

kolyaventuri:
Fake wanted levels are psossible, however I would rather do that as a follow up to this PR. I want to keep this PR scoped purely to the **names** of the effects. Anything else I would consider to be out of scope, and we can do that in a follow up.

ChaosMod/Effects/EffectsInfo.h
pongo1231: I'm not sure why it's required for this effect since it doesn't do any faking and doesn't run any `OnStart` logic that prevents the name from showing up

ChaosMod/Effects/EffectsInfo.h
```diff
@@ -506,13 +507,13 @@ const std::unordered_map<EffectType, EffectInfo> g_effectsMap =
...
 	{EFFECT_PEDS_MINIONS, {"Minions", "peds_minions", true}},
 	{EFFECT_PEDS_MERCENARIES, {"Mercenaries", "peds_mercenaries", true}},
 	{EFFECT_LOOSE_TRIGGER, {"Loose Triggers", "peds_loosetrigger", true}},
 	{EFFECT_MISC_FLAMETHROWER, {"Flamethrowers", "misc_flamethrower", true}},
 	{EFFECT_MISC_DVDSCREENSAVER, {"DVD Screensaver", "misc_dvdscreensaver", true, {}, true}},
-	{EFFECT_PLAYER_FAKEDEATH, {"Fake Death", "player_fakedeath"}},
+	{EFFECT_PLAYER_FAKEDEATH, {"Fake Death", "player_fakedeath", false, {}, false, EffectExecutionType::DEFAULT, "Suicide"}},
```
pongo1231, line 516: The issue is this effect doesn't always play the suicide animation (it can also either do no animation or explode the vehicle the player is in). It should rather be specified in the effect directly so it's able to show a fake name corresponding to the action it's doing (`Suicide` if it's playing the suicide animation, `Explode Current Vehicle` if it's faking a vehicle explosion and nothing if no animation is being played).

Also it's preferable to make it use either the original or (if available) custom name of another effect directly instead of using hardcoded texts (maybe by adjusting the `OverrideEffectName` to take another effect enum item and fetch the name based on that instead).

pongo1231 (changes requested):

PR #327, opened 2021-01-07
[Add Effect] GTA 2

kolyaventuri:
- Adds an extremely disorienting top-down camera view, a la GTA / GTA 2

Demo: https://streamable.com/hk9j7c

### DIFF
--- a/ChaosMod/Effects/db/Player/PlayerGTA2.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerGTA2.cpp
@@ -35,6 +35,7 @@ static void OnTick() {
 static void OnStop() {
 	RENDER_SCRIPT_CAMS(false, true, 500, 0, 1, 0);
 	SET_CAM_AFFECTS_AIMING(camera, true);
+	DESTROY_CAM(camera, true);
 }
 
 static RegisterEffect registerEffect(EFFECT_PLAYER_GTA_2, OnStart, OnStop, OnTick);
--- a/ChaosMod/Effects/db/Player/PlayerGTA2.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerGTA2.cpp
@@ -1,0 +1,11 @@
+#include <stdafx.h>
+
+/*
+* Effect by kolyaventuri
+*/
+
+static void OnStart() {}
+
+static void OnStop() {}
+
+static RegisterEffect registerEffect(EFFECT_PLAYER_GTA_2, OnStart, OnStop);
--- a/ChaosMod/Effects/db/Player/PlayerGTA2.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerGTA2.cpp
@@ -5,6 +5,8 @@
 */
 
 static const float height = 40.f;
+static const float speedFactor = 1.f + (0.5f / (180.f / 2.236936f));
+static float baseFov;
 static Cam camera;
 
 static void OnStart() {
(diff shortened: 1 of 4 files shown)


Sam-V8:
can the camera be made to zoom out when getting into the car? it doesn't provide enough forward vision like GTA 2 did

SeanMirrsen:
Or, if you're really doing it right, have the camera zoom out further as velocity increases.

kolyaventuri:
@Sam-V8 totally! I've upped it since I took that video and it's much better now, just need to re-record it

SeanMirrsen:
Can't test it myself, but from the video it looks hilarious. Other than the brief kerfuffle when the camera tries to switch when you're run over, it's most excellent. Great job. :D

ChaosMod/Effects/db/Player/PlayerGTA2.cpp
```diff
@@ -0,0 +1,40 @@
...
+	SET_CAM_ROT(camera, -90.f, 0.f, 0.0, 2);
+	RENDER_SCRIPT_CAMS(true, true, 500, 0, 1, 0);
+	baseFov = GET_CAM_FOV(camera);
+
+	SET_CAM_AFFECTS_AIMING(camera, false);
+}
+
+static void OnTick() {
+	Ped player = PLAYER_PED_ID();
+	Vector3 rot = GET_ENTITY_ROTATION(player, 2);
```
Last0xygen, line 27: rot is never used

ChaosMod/Effects/db/Player/PlayerGTA2.cpp
```diff
@@ -0,0 +1,40 @@
...
+	float offset = speed * speedFactor;
+
+	SET_CAM_FOV(camera, baseFov + offset);
+	SET_CAM_ROT(camera, -90.f, 0.f, 0.0, 2);
+}
+
+static void OnStop() {
+	RENDER_SCRIPT_CAMS(false, true, 500, 0, 1, 0);
+	SET_CAM_AFFECTS_AIMING(camera, true);
+}
```
Last0xygen, line 38: you should destroy the cam if its not needed anymore

ChaosMod/Effects/db/Player/PlayerGTA2.cpp
```diff
@@ -0,0 +1,40 @@
...
+static const float speedFactor = 1.f + (0.5f / (180.f / 2.236936f));
+static float baseFov;
+static Cam camera;
+
+static void OnStart() {
+	Ped player = PLAYER_PED_ID();
+	Vector3 rot = GET_ENTITY_ROTATION(player, 2);
+
+	camera = CREATE_CAM("DEFAULT_SCRIPTED_FLY_CAMERA", 1);
+	ATTACH_CAM_TO_ENTITY(camera, player, 0.f, 0.f, height, true);
```
Last0xygen, line 17: This most likely will change the camera position, once the player is not standing up straight, right? like riding on a bike for example (not tested)
maybe fixed, if the last flag is switched to false

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerGTA2.cpp
kolyaventuri: Oh nice catch, must have missed this in cleaning up

ChaosMod/Effects/db/Player/PlayerGTA2.cpp
kolyaventuri: oh quite possibly. I'll play around with that. That might explain why the camera goes all wonky when you get run over, or die

PR #333, opened 2021-01-08
Added Effect 'Noclip Vehicle'

Last0xygen:
Disables collision for the player vehicle to any other vehicle or ped


Zevexia:
Disabling collision for ALL vehicles would be a good effect to add too.

ChaosMod/Effects/db/Vehs/VehsNoCollision.cpp
```diff
@@ -0,0 +1,26 @@
...
+{
+	Ped player = PLAYER_PED_ID();
+	if (IS_PED_IN_ANY_VEHICLE(player, false))
+	{
+		Vehicle playerVeh = GET_VEHICLE_PED_IS_IN(player, false);
+		std::vector<Vehicle> foundVehs = GetNearbyVehicles(player, 50, 100);
+		std::vector<Ped> foundPeds = GetNearbyPeds(player, 50, 100);
+		for (Vehicle veh : foundVehs)
+		{
+			SET_ENTITY_NO_COLLISION_ENTITY(veh, playerVeh, true);
```
kolyaventuri, line 17: Does this reset every frame? Or is there cleanup required?

Last0xygen: Every frame. False for the last parameter would be a lasting effect which i don’t know how to disable again 😅

ChaosMod/Util/Player.h
```diff
@@ -37,3 +37,28 @@ inline void TeleportPlayer(const Vector3& coords, bool noOffset = false)
...
+			Ped ped = handles[index];
+			if (GET_ENTITY_COORDS(ped, false).distanceTo(position) < radius)
+			{
+				found.push_back(ped);
+			}
+		}
+	}
+	delete[] handles;
+	return found;
+}
```
pongo1231, line 64: A much more performant and simple way is to simply use the already existing `GetAllPeds` function (especially since you are checking for the distance of each ped anyways)

ChaosMod/Util/Vehicle.h
```diff
@@ -28,4 +28,30 @@ inline Vehicle CreateTempVehicleOnPlayerPos(Hash model, float heading)
...
+			Vector3 vehPos = GET_ENTITY_COORDS(veh, false);
+			if (vehPos.distanceTo(position) < radius)
+			{
+				foundVehs.push_back(veh);
+			}
+		}
+	}
+	delete[] handles;
+	return foundVehs;
+}
```
pongo1231, line 57: Same applies here, no need to execute a native for that when you have a `GetAllVehs` helper function already

ChaosMod/Util/Player.h
Last0xygen: is it that much more performant to iterate all peds instead of just the ones nearby? 🤔

pongo1231: Considering you are asking the game to collect all the peds, filter out all the nearby ones and return them (with native invocation overhead and all) instead of using the available functions which fetch them directly from game memory and cache them for the current frame, pretty much.

Last0xygen: Touché

pongo1231:
I think it's too situational in its current state (only affects player's vehicle and only does something if the player decides to drive through another entity), also the name might not be too clear for people. Making it so every entity has no collisions with each other and renaming it to something like "No Collisions" might make the effect both broader and more clarifying (in my opinion).

Last0xygen:
> 
> 
> I think it's too situational in its current state (only affects player's vehicle and only does something if the player decides to drive through another entity), also the name might not be too clear for people. Making it so every entity has no collisions with each other and renaming it to something like "No Collisions" might make the effect both broader and more clarifying (in my opinion).

as it has to be run every frame to be disabled after the effect ended, i think this can't be done fast enough sadly... 
I guess i have to find some memoryhack then, otherwise i dont see any method

PR #342, opened 2021-01-11
Weaker Gravity Field effect

juhana:
As a suggestion, I made the Gravity Field much weaker so that objects are slowly crawling towards the player instead of just violently crashing into a ball of debris until the effect ends and the player dies. IMHO this makes for a much more interesting effect as the player needs to dodge the approaching objects.


EliRickard:
Should improve crash probability with this effect on weak pcs.

ProfessorBiddle:
This should be a second effect, maybe rename the original gravity field to "Black Hole"

The original one is fantastic fun, especially in cutscenes

juhana:
Effect is weaker in the latest version.

PR #362, opened 2021-01-16
fix(fake-death): Ensure player cannot die during fake death

### LINKED ISSUE #361: [Bug] Fake death can cause a real death when executed just wrong
**Describe the bug**
When you get a fake death in a vehicle you may explode and blast away, the player is not invincible and can still die at this point, presumably this is the same for the ragdoll after the suicide animation, and although harder to reproduce, it's probably the same for when you just die without anything happening, though you'd need someone shooting at you or something like that to really check this

**To Reproduce**
None. It's inconsistent and dependent on luck for which animation, but see [this clip](https://clips.twitch.tv/SuccessfulStupidSangOpieOP)

kolyaventuri:
fixes #361 

- Adjusts `Fake Death` effect to ensure player cannot die during it


kolyaventuri:
i uh. labeled this wrong.

PR #367, opened 2021-01-18
[New Effect] Nap Time

EliRickard:
Sets time to night, Forces all peds and player to halt vehicle (if in one) and get out, forces all peds (excluding player) to perform sleep animation for the duration of the effect. It's on screen name is "It's Nap Time"

It is timed as a "Long" effect.

(This effect and my Where's the funeral PR are being redone as they were made in the master branch of my fork, which would inhibit further effects creation)

### DIFF
--- a/ChaosMod/Effects/db/PedsNapTime.cpp
+++ b/ChaosMod/Effects/db/PedsNapTime.cpp
@@ -1,0 +1,37 @@
+#include <stdafx.h>
+
+/*
+	Effect by Eli Rickard
+*/
+
+static void OnStop()
+{
+	REMOVE_ANIM_DICT("amb@world_human_bum_slumped@male@laying_on_left_side@idle_a");
+}
+
+static void OnTick()
+{
+	SET_CLOCK_TIME(21, 30, 50);
+
+	for (auto ped : GetAllPeds())
+	{
+		if (IS_PED_IN_ANY_VEHICLE(ped, true))
+		{
+			Vehicle pedVeh = GET_VEHICLE_PED_IS_IN(ped, true);
+			TASK_LEAVE_VEHICLE(ped, pedVeh, 256);
+			BRING_VEHICLE_TO_HALT(pedVeh, 0.1f, 10, 0);
+		}
+	}
+
+	REQUEST_ANIM_DICT("amb@world_human_bum_slumped@male@laying_on_left_side@idle_a");
+
+	for (auto ped : GetAllPeds())
+	{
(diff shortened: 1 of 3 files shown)


ChaosMod/Effects/db/PedsNapTime.cpp
```diff
@@ -0,0 +1,37 @@
...
+	REQUEST_ANIM_DICT("amb@world_human_bum_slumped@male@laying_on_left_side@idle_a");
+
+	for (auto ped : GetAllPeds())
+	{
+		if (!IS_PED_A_PLAYER(ped) && !IS_ENTITY_PLAYING_ANIM(ped, "amb@world_human_bum_slumped@male@laying_on_left_side@idle_a", "idle_b", 3))
+		{
+			TASK_PLAY_ANIM(ped, "amb@world_human_bum_slumped@male@laying_on_left_side@idle_a", "idle_b", 4.0f, -4.0f, -1.f, 1, 0.f, true, true, true);
+		}
+	}
+}
```
kolyaventuri, line 35: This could be collapsed into a single for loop, since each time you're just iterating over each ped anyways

ChaosMod/Effects/db/PedsNapTime.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+	for (auto ped : GetAllPeds())
+	{
+		if (IS_PED_IN_ANY_VEHICLE(ped, true))
+		{
+			Vehicle pedVeh = GET_VEHICLE_PED_IS_IN(ped, true);
+			TASK_LEAVE_VEHICLE(ped, pedVeh, 256);
+			BRING_VEHICLE_TO_HALT(pedVeh, 0.1f, 10, 0);
+		}
+	}
```
kolyaventuri, line 24: Thinking about it for 2 seconds, this could be done in the `OnStart`, unless you want any ped that is spawned in _any_ vehicle to do this? I.e., if you drive a bit and new vehicles spawn

EliRickard: I wanted everyone new that spawned it after the fact, to also be sleeping given that it is a timed effect but i'll fix the Getting of all peds twice, now.

EliRickard:
Removed the double of getting all peds.

EliRickard:
Removed original misplaced file.

kolyaventuri (approved):

kolyaventuri:
Make sure to update the `chaos/ChaosMod.vcxproj` file. Easiest way to do this, would be to create new classes + files using Visual Studio as your IDE, and build the mod, though you can update it manually I suppose.

Strictly speaking it will still _work_ if you don't do so, but the change to actually allow the linker to see it will get pulled into another PR or commit.

EliRickard:
Should be done

PR #368, opened 2021-01-18
[New Effect] Where is the Funeral?

EliRickard:
Effect spawns Romero hearse @ Player Coords
On Screen Name: Where is the funeral?
Not Timed Obviously and makes use of existing veh spawner effect.

(This effect and my Nap Timel PR are being redone as they were made in the master branch of my fork, which would inhibit further effects creation)


kolyaventuri (approved):

pongo1231:
Make sure every word starts with an uppercase in the effect name in both `EffectsInfo.h` and `Effects.cs` to remain consistent with all the other effects

EliRickard:
Fixed to match Capitalisation Scheme.

PR #369, opened 2021-01-18
[Improvement] I'm So Tired

EliRickard:
Adds driving mechanics to I'm So Tired, to simulate actual tired driving.


pongo1231:
This was already implemented with #344.

Issue #279, opened 2020-12-25
[Suggestion] Effect: Potato PC


MrRetr0:
Makes graphics settings really low, changes resolution and makes the game run at 30 fps. 30 second timer.

MoneyWasted:
Possibly update Low Render Distance? Add some grainy time cycle and pause the game every ~50ms to simulate stuttering.

Issue #244, opened 2020-12-12
[Suggestion] Effect: Limit FPS


TheRealAyCe:
Lock the framerate to 15 fps for a while

PiotrKowalski:
I would call it "Console POV" instead of "Limit FPS"

AxmPlays:
> I would call it "Console POV" instead of "Limit FPS"

console fps is 30

PiotrKowalski:
> > 
> 
> console fps is 30
That was a joke. You know. That consoles run poorly.

PR #373, opened 2021-01-18
Added Effect 'Console Experience'

Last0xygen:
Limits FPS to max 15
Fixes https://github.com/gta-chaos-mod/ChaosModV/issues/279 and fixes https://github.com/gta-chaos-mod/ChaosModV/issues/244


ChaosMod/Effects/db/Misc/MiscFpsLimit.cpp
```diff
@@ -0,0 +1,17 @@
+#include <stdafx.h>
+
+/*
+	Effect by Last0xygen
+*/
+
+static void OnTick()
+{
+	int time = 1000 / 15;
```
pongo1231, line 9: Should be `static const`, no need to do it every frame

pongo1231:
I think it should be short timed (since it's a quite annoying effect) and it should rather set the FPS to something like 25 instead of 15 which is too stuttery.

PR #385, opened 2021-01-21
MetaEffects Adjustments

Last0xygen:
- [X] Load Config-Settings for MetaEffects
- [x] Inspect Meta-Effect dispatching while TimerSpeed is active
  * Created Timer for Meta-Effects

### DIFF
--- a/ChaosMod/EffectDispatcher.cpp
+++ b/ChaosMod/EffectDispatcher.cpp
@@ -85,8 +85,6 @@ void EffectDispatcher::UpdateTimer()
 
 		m_timerTimer = currentUpdateTime;
 		delta = 0;
-
-		UpdateMetaEffects();
 	}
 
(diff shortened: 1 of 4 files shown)


SeanMirrsen:
You know, I just noticed. Is it just me or is there no "nothing" effect for metaeffects? It seems like one will always activate at the specified timer cutoff, when there should be some chance that things work normally for a while.

pongo1231:
I think the default meta effect dispatch timer should be changed from every 3 minutes to something like 10 minutes instead. This will guarantee things are normal for a good while before a meta effect kicks in.

SeanMirrsen:
Just having a "normality" effect with a decent weighting will do that too. Once every 10 minutes is a bit too long if it happens all the time, and the idea for metaeffects is to make the chaos more chaotic. Metaeffects should have a chance to fire several times in a row, just as much as they should have a chance of not firing for a while.

PR #390, opened 2021-01-23
[New Effect] Fuel Spill

joshuax8:
**Name of effect**
Fuel Spill

**Duration if timed effect (long or short timed?)**
Long

**Effect Description**
Sets VEHICLE_PETROL_TANK_HEALTH to 0 of all nearby vehicles so the leave a trail of flammable fuel behind.


ChaosMod/Effects/db/Vehs/VehsFuelSpill.cpp
```diff
@@ -0,0 +1,15 @@
...
+	Effect by Joshuax8
+*/
+
+#include <stdafx.h>
+
+static void OnTick()
+{
+	for (auto veh : GetAllVehs())
+	{
+		SET_VEHICLE_PETROL_TANK_HEALTH(veh, 0.f);
```
kolyaventuri, line 11: You could run [`SET_VEHICLE_FUEL_LEVEL`](https://docs.fivem.net/natives/?_0xBA970511) here as well, to ensure it never runs out

joshuax8:
Ok, somehow this doesn't seem to work yet. Starting to suspect the car needs to be damaged first or something. Will need to look into how to troubleshoot this.

ChaosMod/Effects/db/Vehs/VehsFuelSpill.cpp
joshuax8: Thanks for the suggestion! Will have to do some testing how long it takes to run out and if this is needed (once it's working).

joshuax8:
> Ok, somehow this doesn't seem to work yet. Starting to suspect the car needs to be damaged first or something. Will need to look into how to troubleshoot this.

Ok so it doesn't like to be passed `0.f` as health but `100.f` seems to work fine. However it only seems to trigger once a player gets into a car... Not what I had in mind. Trying to fix that next.

joshuax8:
Tested a couple of things to get all cars to leak fuel (without needing player to damage / get in to the car), so far without luck:
```
SET_VEHICLE_HAS_BEEN_DRIVEN_FLAG(veh, true);
SET_VEHICLE_HAS_BEEN_OWNED_BY_PLAYER(veh, true);
SET_VEHICLE_DAMAGE(veh, 0.f, 0.f, 0.33f, 1.f, 100.f, false);
SET_VEHICLE_CAN_LEAK_PETROL(veh, TRUE);
SET_DISABLE_VEHICLE_PETROL_TANK_FIRES(veh, FALSE);
SET_DISABLE_VEHICLE_PETROL_TANK_DAMAGE(veh, FALSE);
```
Suggestions are welcome!

ChaosMod/Effects/db/Vehs/VehsFuelSpill.cpp
joshuax8: `SET_VEHICLE_FUEL_LEVEL` seems to be unavailable in the bundled `vendor/scripthookv/inc/natives.h` unfortunately.

joshuax8: `SET_VEHICLE_FUEL_LEVEL` is a FiveM exclusive (not in game by default) so unable to use it because of that (thanks pongo for the hint).

joshuax8:
Ok, let's just merge it like this, we can always expand it. It was still pretty fun while testing,

Last0xygen:
What’s the difference to „oil leaks“?

joshuax8:
> What’s the difference to „oil leaks“?

@Last0xygen The vehicle runs out of fuel and becomes unusable and it seems a lot more prone to catching fire while driving but it is quite similar, I agree, didn't realize that before now.

Last0xygen:
With „oil leaks“ you also can catch fire of your exhaust throws some flames 😅 it’s not really oil but petrol puddles everyone is leaking 
As it looks to me this is a combination of „kill engine of every vehicle“ and „oil leaks“

joshuax8:
I can see how they could be too similar, yes. I'll close the pull request.

Issue #382, opened 2021-01-21
[Bug] Phone prevents solving the Hacking Puzzle


TheRealAyCe:
https://youtu.be/riFd668i8sE?t=324
Apparently the hacking puzzle controls collide with the phone?

PR #447, opened 2021-01-26
Fix Phone being stuck on screen during "Realistic Hacking" minigame.

MoneyWasted:
As seen in DarkViperAU's steam he got stuck with the phone on his screen being unable to complete the mini game. This pull request should fix that issue by "destroying" the phone when the effect is started.


MoneyWasted:
Ignore the merge pull requests I was updating my fork.

kolyaventuri (approved):

Last0xygen:
Doesn’t this change have softlock-potential, if The phone gets destroyed on an important call?

kolyaventuri:
> Doesn’t this change have softlock-potential, if The phone gets destroyed on an important call?

Yes, however typically during the call the phone screen isn't visible

SeanMirrsen:
The phone menu being up causes the softlock with the hacking minigame, not the call. So while there will still be a risk of softlocking, it will be reduced overall.

Last0xygen:
🤔 are you sure? Things requiring your decision like helping Stacy/ jizzle/ amanda or not

SeanMirrsen:
Well I'm obviously not *sure*, but it seems like it'd be a better option.

Maybe the phone can be force-closed instead of being destroyed, delaying the effect activation until the phone UI is no longer active?

Last0xygen:
Yeah, i‘d personally prefer the effect to be waiting for the phone to be removed by itself rather than forcing it :)

TheRealAyCe:
Just connecting this pull request with the bugreport: https://github.com/gta-chaos-mod/ChaosModV/issues/382

PR #449, opened 2021-01-26
New effect: Spawn Roasting Lamar

juhana:
Spawns Lamar who greets the player. The player character dismisses him and Lamar starts insulting the player. When the effect ends Lamar flips off the player and leaves. The player character expresses confusion.

...it's the meme.

### DIFF
--- a/ChaosMod/Effects/db/Peds/PedsSpawnRoastingLamar.cpp
+++ b/ChaosMod/Effects/db/Peds/PedsSpawnRoastingLamar.cpp
@@ -29,6 +29,7 @@ static void OnStart()
 
 	SET_PED_RELATIONSHIP_GROUP_HASH(lamarPed, relationshipGroup);
 	SET_PED_AS_GROUP_MEMBER(lamarPed, GET_PLAYER_GROUP(PLAYER_ID()));
+	SET_ENTITY_INVINCIBLE(lamarPed, true);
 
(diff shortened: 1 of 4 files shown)


TheRealAyCe:
Could maybe even replace Lamar with a random ped model.

ChaosMod/Effects/db/Peds/PedsSpawnRoastingLamar.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+	Hash relationshipGroup;
+	ADD_RELATIONSHIP_GROUP("_ROASTING_LAMAR", &relationshipGroup);
+	SET_RELATIONSHIP_BETWEEN_GROUPS(0, relationshipGroup, GET_HASH_KEY("PLAYER"));
+	SET_RELATIONSHIP_BETWEEN_GROUPS(0, GET_HASH_KEY("PLAYER"), relationshipGroup);
+
+	Ped playerPed = PLAYER_PED_ID();
+	Vector3 playerPos = GET_ENTITY_COORDS(playerPed, false);
+
+	lamarPed = CreatePoolPed(4, model, playerPos.x, playerPos.y, playerPos.z, GET_ENTITY_HEADING(playerPed));
```
pongo1231, line 21: Might make more sense to use the normal `CREATE_PED` instead considering it gets cleaned up by the effect itself afterwards anyways.

pongo1231:
Lamar should probably be invincible as well since you can still blow him up with explosives.

juhana:
Lamar is now invincible during the roasting.

PR #521, opened 2021-02-02
Added effect: Spawn Mugger

Chankaaa:
This is inspired by the muggers that are hireable in freeroam.


Chankaaa:
New request is here: https://github.com/gta-chaos-mod/ChaosModV/pull/524

PR #524, opened 2021-02-03
Added effect: Spawn Mugger

Chanka0:
This is inspired by the muggers that are hireable in freeroam.

The difference between the original pull request and this one is that I moved the new effect onto it's own branch, making it easier for future development.

### DIFF
--- a/ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
+++ b/ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
@@ -1,0 +1,127 @@
+/*
+	Effect by Chanka
+*/
+
+#include <stdafx.h>
+
+//Credit to Last0xygen for their excellent code
+//Without their example of using enum and while in combination to bypass OnTick, this would be much cruder, which I think this still is crude.
+
+//Inspired by the muggers you can hire in freeroam.
+//Mugger spawns within 1-5 units away from player. Mugger then approaches with a knife to attack the player
+//When the player is hit, the mugger will then flee from the player. If they die, they are marked as no longer needed.
+//The amount of money that can be stolen is limited by SET_PED_MONEY to 0xFFFF.
+
+enum MuggerState {
+	spawn = 0,
+	approach,
+	mugging,
+	fleeing,
+	timeout
+};
+
+static int mode = MuggerState::spawn;
+
+//This has its own function to avoid duplicating code.
+static void RemoveMoney(Ped player, Ped mugger) {
+	int money;
+	int id;
+	char statNameFull[32];
+
+	switch (GET_ENTITY_MODEL(player))
+	{
+	case 225514697: // Michael 
+		id = 0;
+		break;
+	case 2602752943: // Franklin
+		id = 1;
+		break;
+	case 2608926626: // Trevor
+		id = 2;
+		break;
+	}
+
+	sprintf_s(statNameFull, "SP%d_TOTAL_CASH", id);
+	auto hash = GET_HASH_KEY(statNameFull);
+	STAT_GET_INT(hash, &money, -1);
+	if (money >= 50000)
+	{
+		STAT_SET_INT(hash, money - 50000, 1);
+		SET_PED_MONEY(mugger, 50000);
+	}
+	else {
+		STAT_SET_INT(hash, money - money, 1);
+		SET_PED_MONEY(mugger, money);
+	}
+}
+
+static void OnStart() {
+
+	mode = MuggerState::spawn;
+	Ped mugger = NULL;
+
+	while (mode < MuggerState::timeout)
+	{
+		WAIT(0);
+
+		Ped playerPed = PLAYER_PED_ID();
+
+		if (IS_ENTITY_DEAD(mugger, false) && mugger != NULL) {
+			SET_PED_AS_NO_LONGER_NEEDED(&mugger);
+			mugger = NULL;
+			mode == MuggerState::timeout;
+			break;
+		}
+
+		switch (mode) {
+		case MuggerState::spawn:
+			Vector3 spawnLoc = GET_ENTITY_COORDS(playerPed, false);
+			mugger = CreatePoolPed(4, -2051422616, spawnLoc.x, spawnLoc.y+1, spawnLoc.z, 0);
+			SET_PED_HEARING_RANGE(mugger, 9999.f);
+			SET_PED_COMBAT_ATTRIBUTES(mugger, 5, true);
+			SET_PED_COMBAT_ATTRIBUTES(mugger, 46, true);
+			SET_PED_COMBAT_ATTRIBUTES(mugger, 0, true);
+			SET_PED_SUFFERS_CRITICAL_HITS(mugger, false);
+			SET_ENTITY_HEALTH(mugger, 500, 0);
+			GIVE_WEAPON_TO_PED(mugger, -538741184, 1, false, true);
+			mode++;
+
+			if (IS_PED_IN_ANY_VEHICLE(playerPed, false))
+			{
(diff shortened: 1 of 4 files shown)


ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
```diff
@@ -0,0 +1,127 @@
...
+			GIVE_WEAPON_TO_PED(mugger, -538741184, 1, false, true);
+			mode++;
+
+			if (IS_PED_IN_ANY_VEHICLE(playerPed, false))
+			{
+				Vehicle playerV = GET_VEHICLE_PED_IS_IN(playerPed, false);
+				SET_PED_INTO_VEHICLE(playerPed, playerV, -2);
+				SET_PED_INTO_VEHICLE(mugger, playerV, -1);
+				TASK_LEAVE_VEHICLE(playerPed, playerV, 4160);
+				TASK_VEHICLE_DRIVE_WANDER(mugger, playerV, 100, 786988);
```
Last0xygen, line 95: why TASK_VEHICLE_DRIVE_WANDER instead of TASK_SMART_FLEE_PED?

ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
```diff
@@ -0,0 +1,127 @@
...
+			SET_PED_COMBAT_ATTRIBUTES(mugger, 46, true);
+			SET_PED_COMBAT_ATTRIBUTES(mugger, 0, true);
+			SET_PED_SUFFERS_CRITICAL_HITS(mugger, false);
+			SET_ENTITY_HEALTH(mugger, 500, 0);
+			GIVE_WEAPON_TO_PED(mugger, -538741184, 1, false, true);
+			mode++;
+
+			if (IS_PED_IN_ANY_VEHICLE(playerPed, false))
+			{
+				Vehicle playerV = GET_VEHICLE_PED_IS_IN(playerPed, false);
```
Last0xygen, line 91: please rename "playerV"

ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
```diff
@@ -0,0 +1,127 @@
...
+				TASK_VEHICLE_DRIVE_WANDER(mugger, playerV, 100, 786988);
+				RemoveMoney(playerPed, mugger);
+				mode = MuggerState::timeout;
+			}
+			break;
+
+		case MuggerState::approach:
+			WAIT(0);
+			TASK_COMBAT_PED(mugger, playerPed, 0, 16);
+			mode++;
```
Last0xygen, line 104: i would combine "approach" and "muggin", as this is the same goal, "attacking the player"
Also you may need to repeat the task to combat the player, if that gets cleared for some reason

ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
```diff
@@ -0,0 +1,127 @@
...
+				SET_PED_TO_RAGDOLL(playerPed, 5000, 7500, 0, false, false, false);
+				mode++;
+			}
+			break;
+
+		case MuggerState::fleeing:
+			WAIT(10);
+			//Ped alertness is changed to prevent some unintended behavior when mugger flees.
+			SET_PED_ALERTNESS(mugger, 0);
+			TASK_SMART_FLEE_PED(mugger, playerPed, 150, -1, false, false);
```
Last0xygen, line 120: would increase the distance to something ridiculus, after all you want him to disappear

Last0xygen (changes requested):
This effect should not be an "OnStartOnly"-effect.
effects handled only with the "onStart"-Function have a defined end, which will be reached. This effect on the other hand could run forever, until the mugger finally attacks the player. 
If the player gets into a plane and starts flying away, the mugger will try to hunt him down for the rest of his life. (and probably the rest of the game 😅 )
i would prefer an OnTick version. If the time Runs out, the mugger will either stop attacking the player or keep running away (as the duration of -1 in the fleeing-command)

ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
```diff
@@ -0,0 +1,127 @@
...
+			SET_PED_AS_NO_LONGER_NEEDED(&mugger);
+			mugger = NULL;
+			mode == MuggerState::timeout;
+			break;
+		}
+
+		switch (mode) {
+		case MuggerState::spawn:
+			Vector3 spawnLoc = GET_ENTITY_COORDS(playerPed, false);
+			mugger = CreatePoolPed(4, -2051422616, spawnLoc.x, spawnLoc.y+1, spawnLoc.z, 0);
```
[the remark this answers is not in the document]
Chankaaa: That was from an older version where there was a random offset to the mugger's spawn location.

ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
```diff
@@ -0,0 +1,127 @@
...
+			SET_PED_HEARING_RANGE(mugger, 9999.f);
+			SET_PED_COMBAT_ATTRIBUTES(mugger, 5, true);
+			SET_PED_COMBAT_ATTRIBUTES(mugger, 46, true);
+			SET_PED_COMBAT_ATTRIBUTES(mugger, 0, true);
+			SET_PED_SUFFERS_CRITICAL_HITS(mugger, false);
+			SET_ENTITY_HEALTH(mugger, 500, 0);
+			GIVE_WEAPON_TO_PED(mugger, -538741184, 1, false, true);
+			mode++;
```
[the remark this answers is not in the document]
Chankaaa: The cases are in order, and it's quicker to use mode++.

ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
Chankaaa: With TASK_VEHICLE_DRIVE_WANDER you can set how they drive. If a mugger who just stole your car immediately stopped at the next light, that wouldn't make much sense.

Chankaaa:
@Last0xygen 
> This effect should not be an "OnStartOnly"-effect.
> effects handled only with the "onStart"-Function have a defined end, which will be reached. This effect on the other hand could run forever, until the mugger finally attacks the player.
> If the player gets into a plane and starts flying away, the mugger will try to hunt him down for the rest of his life. (and probably the rest of the game 😅 )
> i would prefer an OnTick version. If the time Runs out, the mugger will either stop attacking the player or keep running away (as the duration of -1 in the fleeing-command)

I wanted to avoid using OnTick() to keep it in line with all other "Spawn character" effects and because the mugger could die early and the timer for the effect would still be there, making the OnTick() after the mugger's death redundant (Also I wanted the mugger spawning to be a secret, because he originally spawned farther away). 
As for it's behavior, it won't run forever. The effect will stop working once a new effect is started, and the engine will takeover the mugger's behavior as if he was a normal NPC. 
I forgot to reimpliment the distance fix from an older version of the mod, which in addition to the death check use VDIST2 and compared the player's location to the mugger's.

ChaosMod/Effects/db/Peds/PedsSpawnMugger.cpp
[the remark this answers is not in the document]
Last0xygen: what do you mean with "quicker"? 
yeah its a shorter command, but it really is easier to read for everyone else ;)

Last0xygen:
alright, i understand the "the timer for the effect would still be there"-reason, but then at least a max distance check should be implemented

Last0xygen:
Ping @Chankaaa 
Any progress? :)

Last0xygen:
Closed for no response in a long time

PR #526, opened 2021-02-03
Added Effect: Animorph Player

Chankaaa:
Transforms the player into a random animal out of a pool of 24 possible animals.


ChaosMod/Effects/db/Player/PlayerAnimorph.cpp
```diff
@@ -0,0 +1,60 @@
...
+	while (!STREAMING::HAS_MODEL_LOADED(animalHash)) WAIT(0);
+	SET_PLAYER_MODEL(PLAYER_ID(), animalHash);
+	SET_PED_DEFAULT_COMPONENT_VARIATION(PLAYER_PED_ID());
+	SET_MODEL_AS_NO_LONGER_NEEDED(animalHash);
+	_SET_PLAYER_IS_IN_ANIMAL_FORM(true);
+}
+
+static void OnStop()
+{
+	Cleanup();
```
kolyaventuri, line 49: If nothing else is happening in the `OnStop` method, lets just move everything in `Cleanup` into the body of `OnStop`. Separating it out would make sense in some cases, but here it's a little redundant

kolyaventuri: Or, I see now that you're doing the same thing in `OnTick`.. nevermind, carry on :joy:

ChaosMod/Effects/db/Player/PlayerAnimorph.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+	Player weapons are removed when switching character model so I have to add it all back.
+*/
+
+#include <stdafx.h>
+
+static std::vector<const char*> animalModels = {"a_c_boar", "a_c_cat_01", "a_c_chickenhawk", "a_c_chimp", "a_c_chop", "a_c_cow", "a_c_coyote", "a_c_crow", "a_c_deer", "a_c_hen", "a_c_husky", "a_c_mtlion", "a_c_pig", "a_c_pigeon", "a_c_poodle", "a_c_pug", "a_c_rabbit_01", "a_c_rat", "a_c_retriever", "a_c_rhesus", "a_c_rottweiler", "a_c_seagull", "a_c_shepherd", "a_c_westy"};
+static Hash currentModel;
+static std::vector<Hash> playerWeapons = {};
+static std::vector<int> playerAmmo = {};
```
Last0xygen, line 12: would make more sense to save them in a map than separate

ChaosMod/Effects/db/Player/PlayerAnimorph.cpp
```diff
@@ -0,0 +1,60 @@
...
+	currentModel = GET_ENTITY_MODEL(PLAYER_PED_ID());
+	for (Hash weapon : Memory::GetAllWeapons()) {
+		if (HAS_PED_GOT_WEAPON(PLAYER_PED_ID(), weapon, false)) {
+			playerWeapons.push_back(weapon);
+			playerAmmo.push_back(GET_AMMO_IN_PED_WEAPON(PLAYER_PED_ID(), weapon));
+		}
+	}
+
+	//Removing while HAS_MODEL_LOADED causes a crash
+	Hash animalHash = GET_HASH_KEY(animalModels[GET_RANDOM_INT_IN_RANGE(0, static_cast<int>(animalModels.size()))]);
```
Last0xygen, line 38: please use our g_random
`GET_HASH_KEY(animalModels[g_random.GetRandomInt(0, animalModels.size() - 1)])`

ChaosMod/Effects/db/Player/PlayerAnimorph.cpp
```diff
@@ -0,0 +1,60 @@
...
+	REQUEST_MODEL(animalHash);
+	while (!STREAMING::HAS_MODEL_LOADED(animalHash)) WAIT(0);
```
Last0xygen, line 40: please use brakets and new lines

ChaosMod/Effects/db/Player/PlayerAnimorph.cpp
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	playerWeapons = {};
+	playerAmmo = {};
+}
+
+static void OnStart() 
+{
+	//Prep before model change
+	currentModel = GET_ENTITY_MODEL(PLAYER_PED_ID());
```
Last0xygen, line 29: define "PLAYER_PED_ID()" as a variable

ChaosMod/Effects/db/Player/PlayerAnimorph.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+	Cleanup();
+}
+
+static void OnTick()
+{
+	//Softlock prevention
+	if (IS_PED_DEAD_OR_DYING(PLAYER_PED_ID(), 1)) {
+		Cleanup();
+	}
```
Last0xygen, line 57: this just keeps **some** softlocks from happening and disables the effect even after respawning (effect still running then)

Last0xygen (changes requested):
I'm not really happy with this. Although it may work most of the time i see a big softlock potential here.
biggest:
* if the player switches to another player this would result in the other player not turning back to normal
* once the player dies he will be "normal" again. the effect will still be active at this point, so theres an active effect without any effect

because of that i would close this PR imho

pongo1231:
Yup explicitly changing the player model is a big "no no" since it does frequently cause softlocks / crashes during missions. Will have to decline this one unfortunately.

Issue #465, opened 2021-01-28
[Effect Suggestion] "Gun Game"


ProfessorBiddle:
**Duration if timed effect (long or short timed?)**
If not a timed effect, remove this section

Long effect

**Effect Description**
Description of the effect; what should it do?

Like Gun Game in Call of Duty, you have one weapon, and every time you kill someone you get a different random weapon.

ThatDeathbat:
How would it work? Like what order would the guns go in?

ThatDeathbat:
#467 is one that I thought would be easier to code. And you didn’t specify what order the guns would go in, or if you’d go back a gun if you died. 

I also mentioned this post in #467 before you said it was just a duplicate of this recommendation.

#467 is not a duplicate, it’s inspired by this one.

ThatDeathbat:
In Call Of Duty’s Gun Game, the guns have an order in which they are unlocked.  They are not random.

PR #547, opened 2021-02-09
Added Effect Gun Game

ProfessorBiddle:
Every time you shoot you get a new random gun. #465

### DIFF
--- a/ChaosMod/Effects/db/Player/PlayerGunGame.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerGunGame.cpp
@@ -1,0 +1,34 @@
+
+#include <stdafx.h>
+
+//Effect by ProfessorBiddle
+
+static void OnStart()
+{
+	Ped playerPed = PLAYER_PED_ID();
+	REMOVE_ALL_PED_WEAPONS(playerPed, 0);
+	GIVE_WEAPON_TO_PED(playerPed, GET_HASH_KEY("WEAPON_PISTOL"), 9999, false, true);
+}
+
+static void OnTick()
+{
+	Weapon wepGive;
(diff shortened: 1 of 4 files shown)


ChaosMod/Effects/db/Player/PlayerGunGame.cpp
```diff
@@ -0,0 +1,34 @@
...
+	REMOVE_ALL_PED_WEAPONS(playerPed, 0);
+	GIVE_WEAPON_TO_PED(playerPed, GET_HASH_KEY("WEAPON_PISTOL"), 9999, false, true);
+}
+
+static void OnTick()
+{
+	Weapon wepGive;
+	Ped playerPed = PLAYER_PED_ID();
+	if (IS_PED_SHOOTING(playerPed))
+	{
```
Last0xygen, line 18: wouldnt this be TOO much chaos? GunGame usually means that you have to kill someone to get a new gun

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerGunGame.cpp
OHSHIET: Maybe it'd be better if peds' guns changed too after every shot, because right now as I see it only changes for the player (for the player I think it's still better to change the gun after every shot, because with every kill it actually may become less interesting)

ProfessorBiddle: I would like to do it every time the player kills someone, but I'm not exactly sure how to do that.

I had similar issues making Pacifist.

ProfessorBiddle:
Closing PR for now while I rewrite with the same changes from Pacifist.

Issue #529, opened 2021-02-04
Big bang


anto4440:
**Name of effect**
Big bang

**Effect Description**
This is sort of two effects combined. Sort of: Every loaded asset will be teleported right to the player, and then they will be launched away with enormous force. Kinda like Teleport everything to  player and forcefeild, but i think it makes sense.

TheRealAyCe:
Big Bang :D

anto4440:
You're absolutely right

MoDerPo:
that's a great idea!

anto4440:
Closing this as it has been added. However i have never seen it in game.

TheRealAyCe:
> Closing this as it has been added. However i have never seen it in game.

That's because the Pull Request is still open and not merged in. :(

PR #558, opened 2021-02-09
Added effect Big Bang

ProfessorBiddle:
A sort-of combination between TP everything to player and Forcefield. Makes lots of stuff fly away from you very quickly. #529

### DIFF
--- a/ChaosMod/Effects/db/Player/PlayerBigBang.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerBigBang.cpp
@@ -1,0 +1,92 @@
+#include <stdafx.h>
+
+//Effect by ProfessorBiddle, Based on code from Pongo and Last0xygen
+
+static void OnStart()
+{
+	std::vector<Entity> entities;
+	int maxEntities = 20;
+
+	for (auto ped : GetAllPeds())
+	{
+		if (maxEntities == 10)
+		{
+			break;
+		}
+
+		maxEntities--;
+
+		if (!IS_PED_A_PLAYER(ped) && !IS_ENTITY_A_MISSION_ENTITY(ped))
+		{
+			SET_ENTITY_AS_MISSION_ENTITY(ped, true, true);
+			entities.push_back(ped);
+		}
+	}
+
+	auto playerPed = PLAYER_PED_ID();
+
+	for (auto veh : GetAllVehs())
+	{
+		if (maxEntities == 0)
+		{
+			break;
+		}
+
+		maxEntities--;
+
+		if ((!IS_PED_IN_ANY_VEHICLE(playerPed, false) || veh != GET_VEHICLE_PED_IS_IN(playerPed, false)) && !IS_ENTITY_A_MISSION_ENTITY(veh))
+		{
+			SET_ENTITY_AS_MISSION_ENTITY(veh, true, true);
+			entities.push_back(veh);
+		}
+	}
+
+	auto playerPos = GET_ENTITY_COORDS(playerPed, false);
+
+	for (auto entity : entities)
+	{
+		SET_ENTITY_COORDS(entity, playerPos.x, playerPos.y + 2, playerPos.z, false, false, false, false);
+	}
+
+	WAIT(0);
+
+	for (auto entity : entities)
+	{
+		SET_ENTITY_AS_MISSION_ENTITY(entity, false, false);
+		SET_ENTITY_AS_NO_LONGER_NEEDED(&entity);
+	}
+	WAIT(200);
+
+	for (Ped ped : GetAllPeds())
+	{
+		if (ped != playerPed) {
+			entities.push_back(ped);
+		}
+	}
+
+	for (Vehicle veh : GetAllVehs())
+	{
+		if (!IS_PED_IN_VEHICLE(playerPed, veh, false))
+		{
+			entities.push_back(veh);
+		}
+	}
+
+	for (Entity prop : GetAllProps())
+	{
+		entities.push_back(prop);
+	}
+
+
+	for (Entity entity : entities)
+	{
+		Vector3 entityCoord = GET_ENTITY_COORDS(entity, false);
+		//boom
+		for (int i = 0; i < 30; i++)
+		{
+			APPLY_FORCE_TO_ENTITY(entity, 1, entityCoord.x - playerPos.x, entityCoord.y - playerPos.y, entityCoord.z - playerPos.z, 0, 0, 0, 0, false, false, true, false, true);
+		}
+	}
+}
+
+static RegisterEffect registerEffect(EFFECT_PLAYER_BIG_BANG, OnStart);
--- a/ChaosMod/ChaosMod.vcxproj
+++ b/ChaosMod/ChaosMod.vcxproj
@@ -117,6 +117,7 @@
     <ClCompile Include="Effects\db\Peds\PedsSpawnBallaSquad.cpp" />
     <ClCompile Include="Effects\db\Peds\PedsMinions.cpp" />
     <ClCompile Include="Effects\db\Peds\PedsSpawnAngryJimmy.cpp" />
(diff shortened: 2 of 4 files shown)


ChaosMod/Effects/db/Player/PlayerBigBang.cpp
```diff
@@ -0,0 +1,92 @@
...
+	for (auto ped : GetAllPeds())
+	{
+		if (maxEntities == 10)
+		{
+			break;
+		}
+
+		maxEntities--;
+
+		if (!IS_PED_A_PLAYER(ped) && !IS_ENTITY_A_MISSION_ENTITY(ped))
```
Last0xygen, line 19: this could result in adding less than 10 entities, if the first 10 peds are either an mission entity or player

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
```diff
@@ -0,0 +1,92 @@
...
+		{
+			SET_ENTITY_AS_MISSION_ENTITY(ped, true, true);
```
Last0xygen, line 21: why is this needed?

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
```diff
@@ -0,0 +1,92 @@
...
+	for (auto veh : GetAllVehs())
+	{
+		if (maxEntities == 0)
+		{
+			break;
+		}
+
+		maxEntities--;
+
+		if ((!IS_PED_IN_ANY_VEHICLE(playerPed, false) || veh != GET_VEHICLE_PED_IS_IN(playerPed, false)) && !IS_ENTITY_A_MISSION_ENTITY(veh))
```
Last0xygen, line 37: the vehicle could be checked outside the for-loop so it doesnt have to be checked all the time

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
```diff
@@ -0,0 +1,92 @@
...
+			SET_ENTITY_AS_MISSION_ENTITY(veh, true, true);
+			entities.push_back(veh);
+		}
+	}
+
+	auto playerPos = GET_ENTITY_COORDS(playerPed, false);
+
+	for (auto entity : entities)
+	{
+		SET_ENTITY_COORDS(entity, playerPos.x, playerPos.y + 2, playerPos.z, false, false, false, false);
```
Last0xygen, line 48: this sets them next to the player. shouldnt it be scattered around the player?

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+
+	for (Entity entity : entities)
+	{
+		Vector3 entityCoord = GET_ENTITY_COORDS(entity, false);
+		//boom
+		for (int i = 0; i < 30; i++)
+		{
+			APPLY_FORCE_TO_ENTITY(entity, 1, entityCoord.x - playerPos.x, entityCoord.y - playerPos.y, entityCoord.z - playerPos.z, 0, 0, 0, 0, false, false, true, false, true);
+		}
```
Last0xygen, line 88: why apply the force 30 times per item?

Last0xygen: Also, why for EVERY entity. maybe some distance check (in the loops above) would be suitable

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
ProfessorBiddle: I copied the code directly from TpEverything, I'm not sure.

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
ProfessorBiddle: Meant to put them directly above the player, i'll fix this.

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
ProfessorBiddle: The force wasn't strong enough any other way. I will add a distance check.

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
ProfessorBiddle: fixed

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
ProfessorBiddle: fixed

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
Last0xygen: If the force is not strong enough, you have to increase the vector, i.e. „(entityCoord.x - playerPos.x) * 30

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
```diff
@@ -21,10 +21,16 @@ static void OnStart()
 			SET_ENTITY_AS_MISSION_ENTITY(ped, true, true);
 			entities.push_back(ped);
 		}
+		else
+		{
+			maxEntities++;
```
Last0xygen, line 26: Why not just Substract It inside the if? :)

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
ProfessorBiddle: Ah thanks I will fix that.

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
ProfessorBiddle: Errrr that would work too. Fixed.

ProfessorBiddle:
I am unable to test this until scripthook is updated, if anyone with steam/able to run the last patch wants to test it for me a build can be found here: https://drive.google.com/file/d/1FcipTtr5jxA-nQZZsS1tyvLxOwKiGWDO/view?usp=sharing

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
```diff
@@ -90,11 +90,9 @@ static void OnStart()
 		Vector3 entityCoord = GET_ENTITY_COORDS(entity, false);
 		if (GET_DISTANCE_BETWEEN_COORDS(entityCoord.x, entityCoord.y, entityCoord.z, playerPos.x, playerPos.y, playerPos.z, true) <= 20)
 		{
-			//boom
-			for (int i = 0; i < 30; i++)
-			{
-				APPLY_FORCE_TO_ENTITY(entity, 1, entityCoord.x - playerPos.x, entityCoord.y - playerPos.y, entityCoord.z - playerPos.z, 0, 0, 0, 0, false, false, true, false, true);
-			}
+			
+			APPLY_FORCE_TO_ENTITY(entity, 1, (entityCoord.x - playerPos.x) * 200, (entityCoord.y - playerPos.y) * 200 , (entityCoord.z - playerPos.z) * 400, 0, 0, 0, 0, false, false, true, false, true);
```
Last0xygen, line 91: Thinking of it: I should have provided Functions for it. 
so „Vector3 forceVec = (entityCoord - playerPos) * 200;“ should be possible 🤔

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerBigBang.cpp
ProfessorBiddle: I'll fix this soon

PR #589, opened 2021-02-14
make Bouncy Vehicles bouncier at low speeds

ProfessorBiddle:
Bouncy vehicles is a bit weak at low speeds, it now brutally punishes you for hitting things.

### DIFF
--- a/ChaosMod/Effects/db/Vehs/VehsBouncy.cpp
+++ b/ChaosMod/Effects/db/Vehs/VehsBouncy.cpp
@@ -6,6 +6,7 @@
 
 static void OnTick()
 {
+	float velFactor;
 	for (auto veh : GetAllVehs())
 	{
 		SET_ENTITY_INVINCIBLE(veh, true);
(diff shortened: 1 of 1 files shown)


ChaosMod/Effects/db/Vehs/VehsBouncy.cpp
```diff
@@ -6,20 +6,26 @@
...
 	{
 		SET_ENTITY_INVINCIBLE(veh, true);
 		//SET_VEHICLE_FIXED(veh);
 		if (HAS_ENTITY_COLLIDED_WITH_ANYTHING(veh))
 		{
 			Vector3 vel = GET_ENTITY_VELOCITY(veh);
-			APPLY_FORCE_TO_ENTITY(veh, 0, vel.x * -60.f, vel.y * -60.f, vel.z * -60.f, .0f, .0f, .0f, 0, true, true, true, false, true);
+			if ((vel.x < 10) && (vel.y < 10) && (vel.z < 10))
+			{
+				velFactor = 600.f;
```
pongo1231, line 19: 600 just causes them to fly around way too much. 300 is a more reasonable value imo.

ProfessorBiddle: Good call.

PR #590, opened 2021-02-14
Update Jesus Take The Wheel, spawns car if the player is not in one.

ProfessorBiddle:
I figure Jesus would drive a pink Panto.


Numi43:
If Online vehicles didn't despawn, I would let him drive a pink Vapid Clique instead.

TheChilliPL:
Maybe add those sticker things too?

<details><summary>Show image</summary>

image

</summary>

PR #591, opened 2021-02-16
WebProxy System

DasDarki:
Hey,
This is a feature I just came across when watching a stream with some friends. We wanted to play this game where some of us are against one of us, but Twitch wasn't an option because its to complex. So I started to create as I call it a "Web Proxy" which allows connecting to a NodeJS Server which allows the other party to vote via a web panel instead of the twitch chat.

And this is it :)

Have a nice day


pongo1231:
The issue is there's a lot of code duplication going on. The voting proxy should be implemented as a separate controller in the existing voting proxy (if possible) and the common voting logic in the mod itself should be shared between the existing twitch voting and your voting implementation using inheritance to avoid duplicated code.

DasDarki:
Alright, I'll optimize it tomorrow or so :)

DasDarki:
So @pongo1231 I've outsourced the most things for minimizing the code base duplication. But a bit duplication is still there but I would need to produce more code than to leave it as it is.

pongo1231:
Hey!

Sorry for letting the pull request sit in the dust for this long. While I definitely appreciate the contribution, I'll have to unfortunately decline it since it'd simply be more maintenance work for us (which we simply don't have the resources for right now) and we'd rather focus on working on the already existing Twitch integration. Our plan is to eventually provide an "add-on" interface to the mod's voting system to give external programs the ability to provide integrations with other platforms. This would make something like this possible without it needing to be included with the mod itself.

As mentioned though, I appreciate the effort regardless.

PR #595, opened 2021-02-16
Make effect 'Snow' reduce vehicle grip

ProfessorBiddle:
A little more like driving in real snow now.


pongo1231:
It's too extreme + I'm pretty sure snowy weather (like rainy / stormy weather) is affecting handling already anyways

PR #600, opened 2021-02-17
Improve Pacifist effect to make sense.

ProfessorBiddle:
With the help of Pongo, finally found a way to detect when the player hits someone. This works way better and makes much more sense than the old effect.

### DIFF
--- a/ChaosMod/Effects/db/Player/PlayerPacifist.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerPacifist.cpp
@@ -2,19 +2,40 @@
 //Effect by ProfessorBiddle, requested thousands of times on discord
 static void OnTick()
 {
-	Player player = PLAYER_ID();
+	int hits0, hits1, hits2;
+	STAT_GET_INT(GET_HASH_KEY("SP0_HITS"), &hits0, -1);
+	STAT_GET_INT(GET_HASH_KEY("SP1_HITS"), &hits1, -1);
+	STAT_GET_INT(GET_HASH_KEY("SP2_HITS"), &hits2, -1);
+	
+	WAIT(0);
 
-	Entity target;
+	Ped playerPed = PLAYER_PED_ID();
 
-	STOP_GAMEPLAY_CAM_SHAKING(true);
+	int hits0Post, hits1Post, hits2Post;
 
-	if (GET_ENTITY_PLAYER_IS_FREE_AIMING_AT(player, &target))
+	switch (GET_ENTITY_MODEL(playerPed))
 	{
-		if ((IS_ENTITY_A_PED(target) || IS_ENTITY_A_VEHICLE(target)) && !IS_ENTITY_DEAD(target, false))
+	case 225514697: // Michael 
+		STAT_GET_INT(GET_HASH_KEY("SP0_HITS"), &hits0Post, -1);
+		if (hits0Post > hits0)
 		{
-			Ped playerPed = PLAYER_PED_ID();
 			SET_ENTITY_HEALTH(playerPed, 0, 0);
 		}
+		break;
+	case 2602752943: // Franklin
+		STAT_GET_INT(GET_HASH_KEY("SP1_HITS"), &hits1Post, -1);
+		if (hits1Post > hits1)
+		{
+			SET_ENTITY_HEALTH(playerPed, 0, 0);
+		}
+		break;
+	case 2608926626: // Trevor
+		STAT_GET_INT(GET_HASH_KEY("SP2_HITS"), &hits2Post, -1);
+		if (hits2Post > hits2)
+		{
+			SET_ENTITY_HEALTH(playerPed, 0, 0);
+		}
+		break;
 	}
 }
 static RegisterEffect registerEffect(EFFECT_PLAYER_PACIFIST, nullptr, nullptr, OnTick);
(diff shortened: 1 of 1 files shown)


ChaosMod/Effects/db/Player/PlayerPacifist.cpp
```diff
@@ -2,19 +2,43 @@
 //Effect by ProfessorBiddle, requested thousands of times on discord
 static void OnTick()
 {
-	Player player = PLAYER_ID();
+	int hits0, hits1, hits2;
+	STAT_GET_INT(GET_HASH_KEY("SP0_HITS"), &hits0, -1);
+	STAT_GET_INT(GET_HASH_KEY("SP1_HITS"), &hits1, -1);
+	STAT_GET_INT(GET_HASH_KEY("SP2_HITS"), &hits2, -1);
```
Last0xygen, line 8: wouldnt it make more sense to make it so:
* you hold the current player hash and current hits 
* Check on Tick if the player changed
  * if so, set the current hits to the current player hits 
* if not, compare the current hits to the last saved hits
  * if > last hits -> kill
  * if <= hits -> set to saved hits (may happen after mission restart / loading a savegame?)
* Also, ONLY get the hash within the switch case, the rest is always the same procedure and just duplicates the code

Just my 2 cents

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerPacifist.cpp
ProfessorBiddle: Thanks for the input, I'll do this.

TheRealAyCe:
Are these stats affected by the characters killing someone while NPC AI takes over?

ProfessorBiddle:
> Are these stats affected by the characters killing someone while NPC AI takes over?

I'm not sure, it should just count hits by whatever character the player is controlling.

ChaosMod/Effects/db/Player/PlayerPacifist.cpp
```diff
@@ -2,17 +2,50 @@
...
+	switch (playerHash)
+	{
+	case 225514697: // Michael 
+		hitHash = GET_HASH_KEY("SP0_HITS");
+		break;
+	case 2602752943: // Franklin
+		hitHash = GET_HASH_KEY("SP0_HITS");
+		break;
+	case 2608926626: // Trevor
+		hitHash = GET_HASH_KEY("SP0_HITS");
```
pongo1231, line 21: You are checking for the same stats as above here

ChaosMod/Effects/db/Player/PlayerPacifist.cpp
```diff
@@ -2,17 +2,50 @@
...
+		switch (playerHash)
+		{
+		case 225514697: // Michael 
+			hitHash = GET_HASH_KEY("SP0_HITS");
+			break;
+		case 2602752943: // Franklin
+			hitHash = GET_HASH_KEY("SP0_HITS");
+			break;
+		case 2608926626: // Trevor
+			hitHash = GET_HASH_KEY("SP0_HITS");
```
pongo1231, line 42: Same applies here also

pongo1231 (changes requested):

ChaosMod/Effects/db/Player/PlayerPacifist.cpp
ProfessorBiddle: Whoops, didn't test with the other two characters 📦 
I'll fix this now.

Last0xygen:
Couldn’t you just keep the player and lastHitcount „outside“ of onTick?
Set it on OnStart to hash -1 and count -1, then you only need one STAT_GET_INT inside the onTick 🤷‍♂️
no WAIT(0) and so on, I hope you understand what I mean 😅

ChaosMod/Effects/db/Player/PlayerPacifist.cpp
```diff
@@ -2,17 +2,50 @@
...
+	}
+	STAT_GET_INT(hitHash, &playerHits, -1);
+
+	WAIT(0);
 
-	if (GET_ENTITY_PLAYER_IS_FREE_AIMING_AT(player, &target))
+	//check if player changed
+	playerPed = PLAYER_PED_ID();
+	Hash newPlayerHash = GET_ENTITY_MODEL(playerPed);
+	if (newPlayerHash = playerHash)
```
pongo1231, line 31: `==` instead of `=`

ProfessorBiddle:
> Couldn’t you just keep the player and lastHitcount „outside“ of onTick?
> Set it on OnStart to hash -1 and count -1, then you only need one STAT_GET_INT inside the onTick 🤷‍♂️
> no WAIT(0) and so on, I hope you understand what I mean 😅

I think I do, I'll give it a shot. I'm pretty inexperienced with c++ and programming in general, so thanks for all the help!

Last0xygen:
> 
> 
> > Couldn’t you just keep the player and lastHitcount „outside“ of onTick?
> > Set it on OnStart to hash -1 and count -1, then you only need one STAT_GET_INT inside the onTick 🤷‍♂️
> > no WAIT(0) and so on, I hope you understand what I mean 😅
> 
> I think I do, I'll give it a shot. I'm pretty inexperienced with c++ and programming in general, so thanks for all the help!

if you need any help just pm me on discord ;)

ChaosMod/Effects/db/Player/PlayerPacifist.cpp
```diff
@@ -1,20 +1,42 @@
 #include <stdafx.h>
 //Effect by ProfessorBiddle, requested thousands of times on discord
-static void OnTick()
-{
-	Player player = PLAYER_ID();
 
-	Entity target;
+int playerHits, playerHitsLast;
```
pongo1231, line 4: Make this `static` please

pongo1231 (changes requested):

PR #616, opened 2021-02-19
Add references to all ChaosMod contributors to Realistic Hacking

ProfessorBiddle:
Not really important


MoneyWasted:
I think all the contributors would find this to be a nice little touch. I sure think so.

PR #624, opened 2021-02-20
󠀡󠀡

niclasdriver:
󠀡󠀡


ChaosMod/Effects/db/Misc/MiscFlashbang.cpp
```diff
@@ -0,0 +1,27 @@
...
+		f_alpha--;
+	}
+}
+
+static void OnStop()
+{
+	f_alpha = 510;
+}
+
+static RegisterEffect registerEffect(EFFECT_MISC_FLASHBANG, nullptr, OnStop, OnTick);
```
TheRealAyCe, line 27: Why use OnStop and not OnStart to set f_alpha? Then you'd only need to define 510 once.

ProfessorBiddle:
Looks a lot like the flash grenade effect from Hitman 1-2-3 too

ChaosMod/Effects/db/Misc/MiscFlashbang.cpp
kolyaventuri: ^ +1, I would just instantiate `f_alpha` at the top level, and set it to 510 on an `OnStart`

kolyaventuri:
This doesn't make sense to be a timed effect. I would opt to use `WAIT(0)` and just run the flashbang effect in the `OnStart`. Just use like, a `while (f_alpha > 0) { . . . }` in the `OnStart`, and decrement it there with a `WAIT(0)` at the end of the loop

niclasdriver:
[Optional sound effects](https://www.dropbox.com/s/rmzmfaox1nfz677/Flashbang%20Sound%20Effects.zip?dl=1)

Last0xygen:
🤔 why?

PR #706, opened 2021-03-02
Added Effect "What The..."

ProfessorBiddle:
Basically causes a headache, if you don't already have one from the rest of the mod.


Last0xygen:
Wouldn’t this be a little epileptic? 😐
I remember people complaining about the „random weather“-effect being headache material, so I don’t know if that’s a good idea 😅

pongo1231:
I agree, this isn't really suitable for the reason mentioned above.

PR #749, opened 2021-03-07
Adding Discord Rich Presence

MoneyWasted:
Allows the mod to hook with the Discord Client running your system to display the Current Effect and the Previous Effect active along with the amount of time you have been playing in a session.

Would look something like this...
rpc2


pongo1231:
Please make sure to keep the code style consistent with the rest of the code base (hungarian notation, casing, new lines, etc.). Also adding an option in the config utility to toggle this might not be a bad idea (could keep it turned on by default).

MoneyWasted:
Yea not gonna lie I kind of rushed it. It still works though. I'll make some needed changes and update it to the latest version of the code base .

PR #835, opened 2021-03-20
Multi language support and spanish translation for config & effects

JoseGonz11:
In order to automatically translate in game effects, I had to use custon name property becouse I cant program in c++.


pongo1231:
Thanks for the PR, however this unfortunately isn't something that can be integrated into the mod itself simply due to the additional maintenance that'd put on contributors. Currently the best way to provide translations would be by sharing the effects.ini with translations as custom effect names around, however distribution should be much easier if #1001 gets implemented.

Issue #499, opened 2021-02-01
[Effect Suggestion] Aimbot


TheBatmanReal:
**Name of effect**
Potential name(s) for the effect
Aimbot
**Duration if timed effect (long or short timed?)**
If not a timed effect, remove this section
Long

**Effect Description**
Description of the effect; what should it do?

Gives The Player  Aimbot

anto4440:
Actually smart idea. I believe that i got this one once but didnt think about it. It feels fair that such a thing would happen

TheRealAyCe:
What exactly is an "aimbot" here?

anto4440:
I assume it hits an enemy everytime, in the head

PR #837, opened 2021-03-21
Added Feature 'Aimbot'

Last0xygen:
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/499


ChaosMod/Effects/db/Player/PlayerAimbot.cpp
```diff
@@ -0,0 +1,37 @@
...
+		{
+			REQUEST_PED_VISIBILITY_TRACKING(ped);
+			if (ped != playerPed && IS_TRACKED_PED_VISIBLE(ped) && !IS_PED_DEAD_OR_DYING(ped, true) && IS_PED_HUMAN(ped) && !IS_PED_A_PLAYER(ped) && IS_TRACKED_PED_VISIBLE(ped))
+			{
+				Vector3 pedCoord = GET_ENTITY_COORDS(ped, 0);
+				float distance = GET_DISTANCE_BETWEEN_COORDS(playerCoord.x, playerCoord.y, playerCoord.z, pedCoord.x, pedCoord.y, pedCoord.z, true);
+				if (distance <= maxRange)
+				{
+					Vector3 headVector = GET_PED_BONE_COORDS(ped, 0x796E, 0, 0, 0); // Head
+					SET_PED_SHOOTS_AT_COORD(playerPed, headVector.x, headVector.y, headVector.z, true);
```
juhana, line 30: Could add `TASK_AIM_GUN_AT_ENTITY(playerPed, ped, 1, false);` here to make the character aim at the targets

juhana: although it makes the walking animation a bit janky

ChaosMod/Effects/db/Player/PlayerAimbot.cpp
```diff
@@ -0,0 +1,37 @@
...
+				}
```
juhana, line 31: Since you can only shoot at one target per tick, there should be a `break;` here when it finds the first valid target to shoot.

ConfigApp/Effects.cs
```diff
@@ -305,6 +305,7 @@ public enum EffectType
             EFFECT_MISC_FPS_LIMIT,
             EFFECT_META_NO_CHAOS,
             EFFECT_PEDS_ROASTING,
+			EFFECT_PLAYER_AIMBOT,
```
juhana, line 308: These tabs should be spaces (same thing at the end of the file)

ChaosMod/Effects/EffectsInfo.h
```diff
@@ -559,4 +560,5 @@ const std::unordered_map<EffectType, EffectInfo> g_effectsMap =
 	{EFFECT_MISC_FPS_LIMIT, {"Console Experience", "misc_fps_limit", true, {}, true}},
 	{EFFECT_META_NO_CHAOS, {"No Chaos", "meta_nochaos", true, { EFFECT_META_HIDE_CHAOS_UI }, false, EffectExecutionType::META}},
 	{EFFECT_PEDS_ROASTING, {"Roasting", "peds_roasting", true, {}, true}}
+	{EFFECT_PLAYER_AIMBOT, {"Aimbot", "player_aimbot", true}},
```
juhana, line 563: Doesn't compile as is, the trailing comma should be on the previous line. Same thing in the ConfigApp/Effects.cs file.

juhana (commented):
The effect is pretty fun, but doesn't do anything if the player doesn't already have a weapon equipped. It could either force the player to equip a weapon at the start or give the player a weapon if they don't have one.

ConfigApp/Effects.cs
Powercube7: what the hell am i reading

ChaosMod/Effects/db/Player/PlayerAimbot.cpp
Last0xygen: solving it with a wait here, otherwise you would continue to shoot only one target until its dead, with a WAIT-delay it jsut shoots several people

ChaosMod/Effects/EffectsInfo.h
Last0xygen: thats embarassing 😅

ConfigApp/Effects.cs
Last0xygen: i have no idea, why its misalligned here, its perfectly alligned in VS :|

ChaosMod/Effects/db/Player/PlayerAimbot.cpp
Last0xygen: Removed the Waits again. it resulted in really delayed shots when changing the view.
Changed it to firing max 5 shots per tick, this should be enough 🤷

ChaosMod/Effects/db/Player/PlayerAimbot.cpp
Last0xygen: yeah, the walking is hardly working if the aim-task has been set (i.e. walks forwards even if only sidesteps pressed)
-> Don't want to add that 😅

Issue #818, opened 2021-03-16
[Effect Suggestion] Cruise Control


Numi43:
**Name of effect**
Cruise Control

**Duration if timed effect (long or short timed?)**
Short

**Effect Description**
Vehicle won´t drop below the actual velocity. Similar to No Braking Allowed but if you speed up, you can´t drop below that speed and the car won´t decrease velocity by itself. Only hard crash slows the car to a halt.

Could be potentially incorporated to No Braking Allowed to make it more chaotic.

Last0xygen:
Already implemented with „full acceleration“, „help my w key is stuck“ or „no breaking“

Numi43:
Last0xygen, neither full acceleration, help my w key is stuck, nor no braking allowed is doing what I suggest.

Numi43:
Full acceleration = makes the speed the maximal
Help my W key is stuck = the car will accelerate all the time or character moves forward
No Braking Allowed = Can´t press S for braking, but can break naturally.

This effect suggestion = Can´t press S for braking, can´t break naturally, but can control the speed, thus being inadvertenly different to both Full Acceleration and Help my W key is stuck.

Last0xygen:
I get that „full acceleration“ and the stuck w key is „greatly different“, but the only difference between „no breaking“ and this one is that you don’t decelerate by yourself. This can not be done unless you keep on setting the current velocity programmatically, which would break your „stopping by head on collision“, which would not be „detected“

Numi43:
Is there a way to detect if the position was changed? If so, could it be detected say every second? Because it doesn´t matter if you decelerate at crash or a second after the crash. The eyes barely notice that.

Last0xygen:
that needs to be done for every vehicle thats loaded, every second, which would result in breaking cars quickly accelerating every second 🤔

Numi43:
It would be only for the car currently driven. Not the others.

anto4440:
How about you just change it up? So that when the effect is applied, the current speed of player cannot increase, but otherwise the rest is the same?

Da-Best:
From what i thought cruise control meant was that the speed it always the same and it will lock the players ability to go forwards or backwards and they will stick to the same speed they were going before

anto4440:
Its just a name...

Numi43:
You know how cruise control works irl? You can go faster, but braking disables it. This effect would lock braking though. You can go faster, but not slower.

[...]

PR #838, opened 2021-03-21
Added Effect 'Cruise Control'

Last0xygen:
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/818
Stops player vehicle from deceleration


pongo1231:
Should be short timed to avoid situations where a player is stuck driving against a wall in a mission required vehicle

Last0xygen:
Adjusted

Issue #701, opened 2021-03-02
[Effect Suggestion] My Super Cool Effect Idea


zynophobia:
**Name of effect**
Potential name(s) for the effect
Help My SpaceBar Is Stuck
**Duration if timed effect (long or short timed?)**
If not a timed effect, remove this section
I'd say not very long maybe a minute
**Effect Description**
Description of the effect; what should it do?
Just spams jump

(Basically hear me out I play gta 5 pc when I run I always accidentally press the spacebar to a point I get annoyed at myself for doing it)

anto4440:
First of all, change the name to something other than 'My super cool effect idea'
Second, The layout is really terrible:
**Name of effect:**
Help my spacebar is stuck

**Duration if timed effect (long or short timed?)**
Long (Maybe one minute instead of the usual 90 sec)

**Effect description**
Presses the jump key every frame, making player constantly jump.

Third: The effect isnt all too interesting, just kinda boring.

Da-Best:
this effect is basically useless in cars as cars do not have a jump button so it will only work on foot.

PR #851, opened 2021-03-22
Added Effect 'Jump! Jump!'

Last0xygen:
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/701
Forces player to keep jumping


ConfigApp/Effects.cs
```diff
@@ -305,6 +305,7 @@ public enum EffectType
             EFFECT_MISC_FPS_LIMIT,
             EFFECT_META_NO_CHAOS,
             EFFECT_PEDS_ROASTING,
+			EFFECT_PLAYER_JUMP_JUMP,
```
kolyaventuri, line 308: nit: Whitespace

ConfigApp/Effects.cs
```diff
@@ -574,6 +575,7 @@ public enum EffectType
             {EffectType.EFFECT_MISC_FPS_LIMIT, new EffectInfo("Console Experience", EffectCategory.MISC, "misc_fps_limit", true, true)},
             {EffectType.EFFECT_META_NO_CHAOS, new EffectInfo("No Chaos", EffectCategory.META, "meta_nochaos", true)},
             {EffectType.EFFECT_PEDS_ROASTING, new EffectInfo("Roasting", EffectCategory.PEDS, "peds_roasting", true, true)}
+			{EffectType.EFFECT_PLAYER_JUMP_JUMP, new EffectInfo("Jump! Jump!", EffectCategory.PLAYER, "player_jump_jump", true)},
```
kolyaventuri, line 578: ditto

kolyaventuri (approved):
Looks good to me, aside from the whitespace for consistency

Last0xygen:
Strangely enough, they are correctly identend when checked out

ChaosMod/Effects/db/Player/PlayerJumpJump.cpp
```diff
@@ -0,0 +1,18 @@
...
+#include <stdafx.h>
+static int lastCheck = 0;
+
+static void OnTick()
+{
+	Ped player = PLAYER_PED_ID();
+	if (lastCheck + 500 < GET_GAME_TIMER() &&  !IS_PED_JUMPING(player))
+	{
+		lastCheck = GET_GAME_TIMER();
+		TASK_JUMP(player, false, false, false);
```
juhana, line 14: This could also be `TASK_CLIMB` which does the same thing except it makes the player climb things if possible.

juhana (commented):
This doesn't work very well when the player is in a vehicle (they just phase out of the car and start falling through the ground.) In that case it could apply the Jumpy Vehicles effect to the player's vehicle instead.

PR #854, opened 2021-03-23
New effect: Hands Up!

juhana:
All peds except cops drop the weapon they're holding and put their hands up for 5 seconds.


TheRealAyCe:
That includes the player, right? :D

PR #867, opened 2021-03-25
Added additional effect groups

Last0xygen:
"PEDS" and "WEAPONS"


TheRealAyCe:
PEDS is for Spawning Peds? Imo SPAWN_PEDS (and SPAWN_PROP/VEHICLE/OBJECT) or something similar would be a much better name. I think "spawn" is the word everyone is looking for when wanting to create a spawn effect and just PEDS may be confused with effects that just affect peds, especially since there already IS a "SPAWN" group (but only for objects).

But I'm very happy about the feature! :D

Issue #426, opened 2021-01-25
[Effect Suggestion] Spawn Hostile Juggernaut


ThatDeathbat:
**Name**
Spawn Juggernaut

**Effect Description**
Spawns a juggernaut with ballistic armor and a minigun.

ThatDeathbat:
It’s like Griefer Jesus, but has armor and a Minigun

ProfessorBiddle:
Alternately a positive effect could be 'Give Ballistic Equipment" which
puts heavy armor on you and gives you an infinite ammo machine gun for a
duration like the online ballistic equipment

ThatDeathbat:
> Alternately a positive effect could be 'Give Ballistic Equipment" which puts heavy armor on you and gives you an infinite ammo machine gun for a duration like the online ballistic equipment

I added that as a request

TheRealAyCe:
Oh look it's my suggestion :P https://github.com/gta-chaos-mod/ChaosModV/issues/98

PR #889, opened 2021-03-27
Added Effect 'Spawn Juggernaut'

Last0xygen:
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/426


pongo1231:
Same issue as before here; the enemy is too accurate. Instead of swapping out weapons another solution would be to reduce the accuracy instead (to something like 5 or 10)

ProfessorBiddle:
> Same issue as before here; the enemy is too accurate and does too much damage. Instead of swapping out weapons another solution would be to reduce the accuracy instead (to something like 5 or 10)

Or even 2-3, these things are rough even in online missions with spongy characters and OP weapons

Issue #758, opened 2021-03-09
[Effect Suggestion] Blimp strat


lauke1010:
**Name of effect**
Blimp strat

**Effect Description**
It teleports the player into a blimp that is then tp'ed to the Galileo Observatory like the speedrun strat. Undecided if the player should keep his momentum or not.

anto4440:
That just a 'Teleport to XXX'

ZeliTheZealot:
The player should be given the correct position and momentum such that, in a few seconds, he will be well poised to dive out of the blimp at the correct position to do the actual blimp strat

PR #891, opened 2021-03-27
Added Effect 'Blimp Strats'

Last0xygen:
Executing the Blimp Strats (Spawning blimp at the "perfect" position and exiting it)
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/758


pongo1231:
Might want to pause all game scripts until the blimp has crashed into the observatory (`Hooks::EnableFakeTpHook` and `Hooks::DisableFakeTpHook`)

Last0xygen:
Done 👍

ChaosMod/Effects/db/Player/PlayerBlimpStrats.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <stdafx.h>
+#include "Memory/Hooks/ScriptThreadRunHook.h"
+
+static void OnStart()
+{
+	Hash blimpHash = GET_HASH_KEY("blimp");
+	LoadModel(blimpHash);
+	Hooks::EnableScriptThreadBlock();
+	Vehicle veh = CreatePoolVehicle(blimpHash, -377.276, 1055.06, 340.962, 80);
```
pongo1231, line 13: Shouldn't this be a normal vehicle which should be marked as despawnable afterwards instead?

Last0xygen: Yeah, would be a better idea :D

PR #901, opened 2021-03-29
CrashFix - Invalid GroupEffectCount

Last0xygen:
While the groupeffect count gets set on the first start only, they get reduced every time the mod reads the config-file -> reducing the effect-weight to become negative -> crashing if that weight overweights all other effects


pongo1231:
Good catch!

PR #924, opened 2021-04-01
Fake Crash improvements

ProfessorBiddle:
Add chance for screen to go black, and add variation to the pause and skip.


DestroyerXyz:
How about also sort of lagging sometimes during fake crash?

ProfessorBiddle:
> How about also sort of lagging sometimes during fake crash?

What do you mean?

Last0xygen:
There is already a lag effect before „crashing“

lZevex:
Is it possible to reduce the frequency of the effect?

ProfessorBiddle:
> Is it possible to reduce the frequency of the effect?

you can change the effect weights in the config app

lZevex:
The effect is starting to happen too often for people to fall for it, maybe you could make the frequency of the effect unadjustable?

Last0xygen:
imho drawing a black screen wont be a good addition. it wont convince someone "more", as this usually is not a reaction of the game on crashing

Issue #391, opened 2021-01-23
[Effect Suggestion] Spawn Companion Jesus


Jornker:
**Name of effect**
spawn companion Jesus

**Effect Description**
Spawn Greifer Jesus as a companion.

TheRealAyCe:
@Last0xygen  Can you edit the title please?

Last0xygen:
> @Last0xygen Can you edit the title please?

Nope, don’t have those rights.
@Jornker please adjust the title :)

MoneyWasted:
We are currently not adding anymore "Jesus" themed effects.

PR #946, opened 2021-04-03
Added Feature "Spawn Companion Jesus"

Zima2115:
suggested in #391

I have no idea what the NO_WRITHE config flag is, but the griefer jesus effect had it. In testing, I didn't see a difference with, or without that flag.


SeanMirrsen:
At a guess (from the name), the tag is supposed to affect whether the NPC immediately gets up and starts firing after being knocked down, or 'loiters' on the ground for a bit. Hence 'writhe' - in pain, as it were.

juhana:
Writhing is the animation where peds squirm on the ground for a while and then die. It happens when their health gets below a certain level so setting no writhe prevents them from dying too early.

I don't think this effect is a very good addition, Jesuses are so iconic to the mod that they should always be griefers IMHO.

Zima2115:
I mean, there already is a driver Jesus. And the moment of confusion whether the Jesus that just spawned is friendly or not adds to the chaos imo.

And thanks for the explanation!

Last0xygen:
I think i have to agree with @juhana 
The addition of "jesus but friendly" is not a good addition to the mod, sorry

Issue #958, opened 2021-04-04
[Effect Suggestion] Tank Bois


Salzdose:
**Name of effect**
Tank Bois

**Duration if timed effect (long or short timed?)**
Immediatly

**Effect Description**
Same as Bus Bois, but it puts all NPCs into Rhinos

TheRealAyCe:
"Tanks alot"

anto4440:
i am confident the engine canty handle that.

juhana:
If it can handle Bus Bois, why couldn't it handle this?

anto4440:
it could handle bus bois. it crashed in the beginning.

Den tir. 6. apr. 2021 kl. 17.55 skrev Juhana Leinonen <
***@***.***>:

juhana:
Yes, and the issue was fixed and it doesn't crash anymore.

anto4440:
By changing it so that everything doesnt spawn in at once. If you actually
were active on the platform, i wouldnt have to babysit you like this.

Den tir. 6. apr. 2021 kl. 18.14 skrev Juhana Leinonen <
***@***.***>:

juhana:
Ok, sorry. I thought that when the suggestion said "same as Bus Bois" it meant "same as the current version of Bus Bois" but apparently it's "same as the old version of Bus Bois that crashed the game" instead.

Last0xygen:
> By changing it so that everything doesnt spawn in at once. If you actually were active on the platform, i wouldnt have to babysit you like this.

???



> Ok, sorry. I thought that when the suggestion said "same as Bus Bois" it meant "same as the current version of Bus Bois" but apparently it's "same as the old version of Bus Bois that crashed the game" instead.

It’s a valid request, we‘ll see if it gets implemented

anto4440:
> Ok, sorry. I thought that when the suggestion said "same as Bus Bois" it meant "same as the current version of Bus Bois" but apparently it's "same as the old version of Bus Bois that crashed the game" instead.

Yeah thats fair, my bad. I had a though day yesterday, sorry for letting it out on you.

ntill:
I feel like this would be pretty funny with the "Invade And Persuade" tank instead of the Rhino.  Would prevent a lot of collision issues.

JustTunerYT:
[...]

PR #982, opened 2021-04-06
Added Effect 'Tanks A Lot'

Last0xygen:
Same as 'Bus Bois' but with tanks lol
https://github.com/gta-chaos-mod/ChaosModV/issues/958


ProfessorBiddle:
oh no

i love it

OHSHIET:
Maybe it should have its own name, something like "Tank Fellas", or anything else you can think of

SeanMirrsen:
I like "Tanks A Lot". Or possibly "Blitzkrieg". :P

ProfessorBiddle:
> Maybe it should have its own name, something like "Tank Fellas", or anything else you can think of

Yeah, with Scooter Brothers and Bus Bois, a rename would be a good idea. 

A few ideas from thesaurus.com:
'Tank Chums'
'Tank Buds'
'Tank Kinsmen'
'Tank Chaps'
'Tank Blokes'
'Tank Homies'
'Tank Dudes'

ChaosMod/Effects/db/Peds/PedsTankBois.cpp
```diff
@@ -0,0 +1,60 @@
...
+			}
+		}
+	}
+}
+
+static RegisterEffect registerEffect(EFFECT_PEDS_TANK_BOIS, OnStart, nullptr, nullptr, EffectInfo
+    {
+        .Name = "Tanks A Lot",
+        .Id = "peds_tank_bois",
+        .EffectGroupType = EffectGroupType::SPAWN_GENERIC
```
pongo1231, line 58: Not sure if I'd classify it as a "generic spawn", if anything this might need a new effect group (though not sure if we need one with 3 similar effects just yet).

ProfessorBiddle: > not sure if we need one with 3 similar effects just yet

There are a whole bunch of suggestions for similar effects, an effect group might not be a bad idea

PR #995, opened 2021-04-07
Improve Heavy Recoil

ProfessorBiddle:
Still really irritating, but smoother and less jerky. More of a COD feel now, but more extreme.


anto4440:
This is confusing me a little; Heavy recoil is the effect that makes you go the opposite direction of where youre shooting, right?

ProfessorBiddle:
> This is confusing me a little; Heavy recoil is the effect that makes you go the opposite direction of where youre shooting, right?

it pushes your gun upwards and randomly side to side while shooting.

PR #996, opened 2021-04-07
Remove Pacifist Workaround with Explosives

ProfessorBiddle:
Removes explosive weapons so that you can't get around Pacifist. A bit mean, but the whole effect is mean.


Last0xygen:
Not needed anymore, thats already working now :)

Issue #1026, opened 2021-04-12
[Bug] "Teleport Player A Few Meters" can teleport player under map


joshuax8:
**Describe the bug**
Teleport Player A Few Meters can teleport player under map (no Z-axis check).

**To Reproduce**
Execute Teleport Player A Few Meter a couple of times.

**Crash Dump (if a game crash)**
N/A

I'll open a merge request to fix this.

PR #1027, opened 2021-04-12
Added GET_GROUND_Z_FOR_3D_COORD to "Teleport Player A Few Meters" effect.

joshuax8:
See issue #1026

### DIFF
--- a/ChaosMod/Effects/db/Player/PlayerTpController.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerTpController.cpp
@@ -158,9 +158,9 @@ static void OnStartFront()
 	Ped playerPed = PLAYER_PED_ID();
 	Vector3 newPos = GET_OFFSET_FROM_ENTITY_IN_WORLD_COORDS(playerPed, 0.f, 50.f, 0.f);
 	float groundZ;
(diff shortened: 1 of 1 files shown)


ChaosMod/Effects/db/Player/PlayerTpController.cpp
```diff
@@ -157,8 +157,10 @@ static void OnStartFront()
 {
 	Ped playerPed = PLAYER_PED_ID();
 	Vector3 newPos = GET_OFFSET_FROM_ENTITY_IN_WORLD_COORDS(playerPed, 0.f, 50.f, 0.f);
+	float groundZ;
+	GET_GROUND_Z_FOR_3D_COORD(newPos.x, newPos.y, 1000.f, &groundZ, false, false);
```
Last0xygen, line 161: you should check, if it succeeded and use "newPos.z" as a fallback

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerTpController.cpp
joshuax8: Added the check!

Last0xygen: 👌

PR #1029, opened 2021-04-12
Added a "Fake Detonate Current Vehicle"

terajoellang:
Added a "Fake Detonate Current Vehicle" (Maybe someone finds a better name for that).

I think it will be pretty funny, since the player might jump out and then nothing happens and then he sees that it was just a fake detonation.

(this is my first time contributing, so I am sorry if I forgot something or did something that is against the rules. If there are rules, pls let me know)

Also I am not yet 100% happy with the code since it contains some code from the real effect and therefore a big part of the code could maybe be put somewhere else and just check if it is fake or not and therefore execute some lines or not execute some lines. But I haven't really found that approach yet in the repo (haven't really searched a lot though), so maybe this is not a practice that is being persued here, so maybe everything is already fine.


terajoellang:
Rename possibilities: Fake Vehicle Detonation or Fake Current Vehicle Detonation or Fake Detonation of Current Vehicle

Da-Best:
isnt this already in "fake death" if you are in a car or am i miss understanding it

SeanMirrsen:
> isnt this already in "fake death" if you are in a car or am i miss understanding it

"Detonate" is a new effect, and is different. It gives you a few seconds of warning with a steadily intensifying beeping noise before the car blows, giving you time to jump out.

This, as I understand, just gives the beeping.

terajoellang:
Exactly, this just gives the beeping and makes the player freak out and maybe jump out of the vehicle just to realize that it was fake afterwards ;)

SeanMirrsen:
Also I'd recommend waiting for #325 to be merged, and implementing the changes from that into this. It will be much more convincing if the effect name doesn't immediately give it away.

terajoellang:
Well, currently the „Detonate Current Vehicle“ and „ Fake Detonate Current Vehicle“ are solely handled inside of the OnStart Function. The OnTick Function is not used (which is wrong, but I am pretty sure, that this approach was used as a workaround in order to not show the Effect Name until it is over. Therefore it could be merged before #325 since the effect name is not shown until the explosion starts in both the real and fake effect.

However, I have not tested this effect yet (I don‘t have any experience with GTA 5 Modding and would first have to set everything up), but if someone could test it and confirm that it works as intended that would be great and then it could be merged.

SeanMirrsen:
Actually I was watching DarkViper's stream, and... apparently if you leave the vehicle when Detonate activates, it doesn't explode? It's kind of underwhelming. :P

terajoellang:
Yeah I was also confused, but I thought maybe that‘s intended in order for the player to also be able to save the vehicle and not only the character. But maybe it is not intended like that, would be pretty easy to fix though. But I don’t feel like I am in a position to decide that, what do you guys think?

SeanMirrsen:
Suggestion! If you can squeeze it in somehow:

When the Fake Detonate timer runs out, if the player is still in the vehicle (because they didn't realize, or couldn't leave the mission vehicle or any other reason), give the player a stack of C4.

The bomb didn't go off, why waste it? :P

terajoellang:
Hmmm, but we can't really let the player know that he just got that C4 (except by displaying a message, but I think this would go too far), so it is a bit pointless if the player doesn't know that he just got a stack of C4. Of course he would probably be surprised at some point, but idk, I don't think it's necessary to add that, sorry, but thanks for the idea :)

Last0xygen:
This is already part of the "Fake Death"-Effect currently. with a certain percentage the 'Detonate Current Vehicle' beeping starts and 'fake explodes' the car after the beeping ends

PR #1037, opened 2021-04-12
Added Effect "The Blade Hungers"

ProfessorBiddle:
This is (sort of) from an escalation in Hitman 3 where 47 is given a katana and told "The Blade Hungers"

The screen flickers and slowly fades to red, and if it goes fully red you die. To reset the fade, you have to kill someone with the machete that the player is forced to hold.


BillNyeTheRussianSpyV2:
Yes. I loved that part of the esclation

ChaosMod/Effects/db/Player/PlayerBladeHunger.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+static void OnTick()
+{
+	Ped playerPed = PLAYER_PED_ID();
+
+	for (Hash weapon : Memory::GetAllWeapons())
+	{
+		if (weapon != bladeHash)
+		{
+			REMOVE_WEAPON_FROM_PED(playerPed, weapon);
```
Last0xygen, line 29: Why remove all weapons? You can just force equip a machete and disable weapon switching

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerBladeHunger.cpp
ProfessorBiddle: I didn't think about that at all, that's a good idea lol

Last0xygen: Now add blocking any weapon-change 🤪

ProfessorBiddle: No need, the equipnow tag forces the player to equip the machete at all times

Last0xygen: sure, but why let him? ;P yeah, but looks good to me 👍

pongo1231:
You might want to consider making it short timed to make sure it doesn't go on for too long (or moving the logic to OnStart + a while loop and stopping the effect after a (couple of) kill(s) / player death).

ProfessorBiddle:
>(or moving the logic to OnStart + a while loop and stopping the effect after a (couple of) kill(s) / player death).

I'll give this a try, but as a short effect it doesn't really work well as the time it takes to do anything is about 20-25 seconds

pongo1231:
In that case I'll just merge it in its current state and then either keep or change it depending on the feedback it gets.

Issue #1039, opened 2021-04-12
[Bug] Set player into random vehicle seat - why?


TheRealAyCe:
https://www.twitch.tv/videos/984800744?t=0h28m30s ~0:28:30
https://www.twitch.tv/videos/984800744?t=0h40m50s ~0:40:50

Why does this happen? Can it be fixed?

TheRealAyCe:
https://github.com/gta-chaos-mod/ChaosModV/issues/730 duplicate of this, i will close the other one

Last0xygen:
https://github.com/gta-chaos-mod/ChaosModV/pull/1040

PR #1040, opened 2021-04-12
Check if vehicle is above ground and has collisions before TP.

joshuax8:
This pull request adds code that loops through random vehicles and checks if they are above the ground and have collisions. This addition should improve situations like the examples provided by another user in issue #1039.

Note: There is still a situation where the player gets set in a vehicle which is invisible and de-spawns quickly after being set into it. I've not been able to find a check to skip those vehicles yet but this pull request should already improve things.


ChaosMod/Effects/db/Player/PlayerRandomVehSeat.cpp
```diff
@@ -23,7 +23,19 @@ static void OnStart()
 		{
 			Ped playerPed = PLAYER_PED_ID();
 
-			Vehicle veh = vehs[g_random.GetRandomInt(0, vehs.size() - 1)];
+			Vehicle veh;
+			for (int i = 0; i < 100; i++)
```
ProfessorBiddle, line 27: Why 100? shouldn't this loop until a suitable vehicle is found?

joshuax8: It's more or less a failsafe so it won't loop forever if something goes wrong. During testing I haven't seen it loop more than 5 times to find a suitable vehicle tbh.

ProfessorBiddle (approved):

ChaosMod/Effects/db/Player/PlayerRandomVehSeat.cpp
terajoellang: Yeah we don't want it to get stuck when "no traffic" is active ;)
Instead of 100 we could take the vehs.size() though.

joshuax8: Good idea, see new commit!

juhana: The problem now is that if there's a very small amount of valid vehicles there's a high chance that it won't find them. For example if there's one valid vehicle and one invalid vehicle there's a 25% chance that it checks the invalid vehicle both times and never finds the valid one. Something like 10 loops should be fine for performance and not missing anything. Another option is to remove invalid vehicles from the vector when they're detected, then the loop could be equal to the vector size with no chance of missing valid vehicles.

Also the current code loops one time fewer than the amount of vehicles so it's going to miss valid vehicles even more often.

Last0xygen: I would move that check to the top, checking it every vehicle before moving it to the array of possible vehicles. This way you are sure to have a valid vehicle before selecting a random one

joshuax8: Thanks for the feedback everyone. Pushed a new commit moving the checks to the vehicle array filling code.

ChaosMod/Effects/db/Player/PlayerRandomVehSeat.cpp
```diff
@@ -14,15 +14,21 @@ static void OnStart()
 
 		std::vector<Vehicle> vehs;
 
+		float groundZ;
 		for (Vehicle veh : GetAllVehs())
 		{
-			vehs.push_back(veh);
+			Vector3 vehPos = GET_ENTITY_COORDS(veh, false);
+			bool useGroundZ = GET_GROUND_Z_FOR_3D_COORD(vehPos.x, vehPos.y, vehPos.z, &groundZ, false, false);
+			bool vehCollision = HAS_COLLISION_LOADED_AROUND_ENTITY(veh);
```
Last0xygen, line 22: Combine these bools please. The other check doesn’t have to be executed if the first check already failed
Same thing for the other effect

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerRandomVehSeat.cpp
joshuax8: I'm not an actual dev I must admit but I gave it a shot (see new commit) 👍

ChaosMod/Effects/db/Player/PlayerRandomVehSeat.cpp
```diff
@@ -18,9 +18,7 @@ static void OnStart()
 		for (Vehicle veh : GetAllVehs())
 		{
 			Vector3 vehPos = GET_ENTITY_COORDS(veh, false);
-			bool useGroundZ = GET_GROUND_Z_FOR_3D_COORD(vehPos.x, vehPos.y, vehPos.z, &groundZ, false, false);
-			bool vehCollision = HAS_COLLISION_LOADED_AROUND_ENTITY(veh);
-			if (useGroundZ && vehCollision)
+			if ((GET_GROUND_Z_FOR_3D_COORD(vehPos.x, vehPos.y, vehPos.z, &groundZ, false, false)) && HAS_COLLISION_LOADED_AROUND_ENTITY(veh))
```
Last0xygen, line 21: almost 😅 too many brackets :)
`if (GET_GROUND_Z_FOR_3D_COORD(vehPos.x, vehPos.y, vehPos.z, &groundZ, false, false) && HAS_COLLISION_LOADED_AROUND_ENTITY(veh))`

ChaosMod/Effects/db/Player/PlayerSetIntoRandomVeh.cpp
```diff
@@ -12,9 +12,7 @@ static void OnStart()
 	for (Vehicle veh : GetAllVehs())
 	{
 		Vector3 vehPos = GET_ENTITY_COORDS(veh, false);
-		bool useGroundZ = GET_GROUND_Z_FOR_3D_COORD(vehPos.x, vehPos.y, vehPos.z, &groundZ, false, false);
-		bool vehCollision = HAS_COLLISION_LOADED_AROUND_ENTITY(veh);
-		if (veh != playerVeh && useGroundZ && vehCollision)
+		if (((veh != playerVeh) && (GET_GROUND_Z_FOR_3D_COORD(vehPos.x, vehPos.y, vehPos.z, &groundZ, false, false))) && HAS_COLLISION_LOADED_AROUND_ENTITY(veh))
```
Last0xygen, line 15: here too, no need to make additional brackets if you compare everything with AND
-> `if (veh != playerVeh && GET_GROUND_Z_FOR_3D_COORD(vehPos.x, vehPos.y, vehPos.z, &groundZ, false, false) && HAS_COLLISION_LOADED_AROUND_ENTITY(veh))`

Last0xygen (changes requested):

Last0xygen (approved):

PR #1053, opened 2021-04-13
Added effect "Eternal Screams"

### LINKED ISSUE #631: [Effect Suggestion] Screaming Contest
**Name of effect**
(Unsure about other names... uh,) Can't Stop Screaming

**Duration if timed effect (long or short timed?)**
10-15 Seconds

**Effect Description**
Make all nearby NPC's and Player's character do random screaming until the time duration is over.
(Screaming as in, type of screaming that a NPC/Player character would sound when falling)

### LINKED ISSUE #1052: [Effect Suggestion] Eternal Screams
**Name of effect**
Eternal Screams

**Duration if timed effect (long or short timed?)**
I think short is annoying enough :)

**Effect Description**
All peds start screaming like they are on fire.
Demo: https://www.youtube.com/watch?v=1QkTWKnft_Q

joshuax8:
Pull request to close #1052 (Would also close #631).


ConfigApp/Effects.cs
```diff
@@ -588,6 +589,7 @@ public enum EffectType
             {EffectType.EFFECT_PLAYER_AIMBOT, new EffectInfo("Aimbot", EffectCategory.PLAYER, "player_aimbot", true)},
             {EffectType.EFFECT_PEDS_SPAWN_BIKER, new EffectInfo("Spawn Biker", EffectCategory.PEDS, "peds_spawn_biker")},
             {EffectType.EFFECT_PEDS_SPAWN_SPACE_RANGER, new EffectInfo("Spawn Space Ranger", EffectCategory.PEDS, "peds_spawn_space_ranger")},
+            {EffectType.EFFECT_PEDS_ETER_SCREAMS, new EffectInfo("Eternal Screams", EffectCategory.PEDS, "peds_eter_screams")},
```
Last0xygen, line 592: id does not match the "correct" one  -> "peds_eternal_screams"
Please also check all other values you created, they're mostly "ETER_SCREAMS" or PedsEterScreams.cpp :)

Last0xygen (changes requested):

ConfigApp/Effects.cs
joshuax8: Ok, like this, right?

Last0xygen (approved):

ConfigApp/Effects.cs
Last0xygen: Perfect 👌

ProfessorBiddle:
Playing a facial animation to make them look like they are screaming might be a nice touch :)

ChaosMod/Effects/db/Peds/PedsEternalScreams.cpp
```diff
@@ -0,0 +1,29 @@
...
+	{
+		PLAY_PAIN(ped, 8, 0, 0);
+	}
+}
+
+static void OnStop()
+{
+	for (Ped ped : GetAllPeds())
+	{
+		DISABLE_PED_PAIN_AUDIO(ped, true);
```
pongo1231, line 19: Any reason to completely disable any pain sounds for every ped on effect end?

ConfigApp/Effects.cs
```diff
@@ -588,6 +589,7 @@ public enum EffectType
             {EffectType.EFFECT_PLAYER_AIMBOT, new EffectInfo("Aimbot", EffectCategory.PLAYER, "player_aimbot", true)},
             {EffectType.EFFECT_PEDS_SPAWN_BIKER, new EffectInfo("Spawn Biker", EffectCategory.PEDS, "peds_spawn_biker")},
             {EffectType.EFFECT_PEDS_SPAWN_SPACE_RANGER, new EffectInfo("Spawn Space Ranger", EffectCategory.PEDS, "peds_spawn_space_ranger")},
+            {EffectType.EFFECT_PEDS_ETERNAL_SCREAMS, new EffectInfo("Eternal Screams", EffectCategory.PEDS, "peds_eternal_screams")},
```
pongo1231, line 592: Make sure to also add the corresponding "timed" flags here

pongo1231 (changes requested):

pongo1231:
Also you should consider making it short timed (since I could see it being a bit too annoying to listen to for 90 seconds :))

ChaosMod/Effects/db/Peds/PedsEternalScreams.cpp
pongo1231: Actually just noticed it isn't registered in the `RegisterEffect` declaration so it's definitely redundant

pongo1231:
I just made the proposed changes myself. Hope it isn't an issue, though feel free to comment if so.

PR #1064, opened 2021-04-14
'Spawn Bus' fills bus with peds after it spawns

ProfessorBiddle:


anto4440:
I missed the part where that's a problem

TheRealAyCe:
It's not a problem, this PR makes it so that the bus is filled, which is marvellous.

Issue #1062, opened 2021-04-14
[Effect Suggestion] Fling player


FAR95d:
**Name of effect**
Fling player

**Effect Description**
Its like launch player but horizontal... It throws the player in a random direction... If it could it could be everything...(Btw idk where to check to see if it was suggested or not)

anto4440:
Good idea!

PR #1077, opened 2021-04-15
Added Effect 'Fling Player'

Last0xygen:
Launches the player into a random direction
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/1062


ProfessorBiddle:
may be good to replace 'Launch Player Up'

PR #1214, opened 2021-04-28
Added effect "Innocence Is Illegal"

ShySkream:
Performing crimes and hurting people causes your wanted level to decrease instead of increase, while not doing so for a period of time will make it go up repeatedly.


ChaosMod/Effects/db/Player/PlayerIllegalInnocence.cpp
```diff
@@ -0,0 +1,140 @@
...
+	Player player = PLAYER_ID();
+	int wantedLevel = PLAYER::GET_PLAYER_WANTED_LEVEL(player);
+
+	Hash playerHash = GET_ENTITY_MODEL(PLAYER_PED_ID());
+	#pragma endregion
+
+	#pragma region check if player ran over someone
+	if (GET_TIME_SINCE_PLAYER_HIT_PED(player) < 200 && GET_TIME_SINCE_PLAYER_HIT_PED(player) != -1) {
+		lastWantedLevel = 0;
+		SET_PLAYER_WANTED_LEVEL(player, lastWantedLevel, false);
```
ProfessorBiddle, line 47: you should extract this and call it each time you use it

ProfessorBiddle (changes requested):

ChaosMod/Effects/db/Player/PlayerIllegalInnocence.cpp
ShySkream: This would make the code a little shorter, but it would also result in a lot of extra calls when nothing is happening.  Not that the extra calls would have much impact, I just like it to be efficient.

ProfessorBiddle: > This would make the code a little shorter, but it would also result in a lot of extra calls when nothing is happening. Not that the extra calls would have much impact, I just like it to be efficient.

i think i selected the wrong bit, i mean just the lines that set the wanted level, not the whole if statement

ShySkream: > i think i selected the wrong bit, i mean just the lines that set the wanted level, not the whole if statement

I was referring to the wanted level lines as well, yes.

PR #1237, opened 2021-04-30
Added effect "Zoom Zoom Cam"

Mo-xii:
Zoom in and out camera effect.

This is my very first time doing anything in C++ or with GitHub as a whole. So I still have plenty to learn.
I'm sure you'll have a better idea on how to create that effect.


ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
```diff
@@ -0,0 +1,72 @@
+/*
+    Effect by Moxi, based on code from DrUnderscore & ProfessorBiddle
+*/
+
+#include <stdafx.h>
+static Camera zoomCamera = 0;
+float camZoom = 80.f;
```
ProfessorBiddle, line 7: these should be static. Otherwise this looks really good :)

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
```diff
@@ -0,0 +1,72 @@
...
+const float camZoomRate = 0.6f;
+int n = 0;
+
+static void UpdateCamera()
+{
+    auto coord = CAM::GET_GAMEPLAY_CAM_COORD();
```
Last0xygen, line 13: Please use the corresponding typename instead of „auto“
Namespace prefixes like „CAM::“ can be omitted

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
```diff
@@ -0,0 +1,72 @@
+/*
+    Effect by Moxi, based on code from DrUnderscore & ProfessorBiddle
+*/
+
+#include <stdafx.h>
+static Camera zoomCamera = 0;
+float camZoom = 80.f;
+const float camZoomRate = 0.6f;
+int n = 0;
```
Last0xygen, line 9: This seems to be a flag whether it should decrease or increase
I would go another way:
* keep a variable which holds the amount it should change (already got the base with „ camZoomRate“)
* once you get below or above the limits -> negate the amount („* -1“)
* This way you don’t need to keep track of you’re increasing or decreasing and can bounds-check with one if and one negate statement

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto coord = CAM::GET_GAMEPLAY_CAM_COORD();
+    auto rot = CAM::GET_GAMEPLAY_CAM_ROT(2);
+    CAM::SET_CAM_PARAMS(zoomCamera, coord.x, coord.y, coord.z, rot.x, rot.y, rot.z, camZoom, 0, 1, 1, 2);
+}
+
+static void OnStart()
+{
+    zoomCamera = CAM::CREATE_CAM("DEFAULT_SCRIPTED_CAMERA", 1);
+    CAM::RENDER_SCRIPT_CAMS(true, true, 10, 1, 1, 1);
+}
```
Last0xygen, line 22: „CamZoom“ should be reset to the initial 80 here

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
```diff
@@ -0,0 +1,72 @@
...
+        {
+            camZoom += camZoomRate;
+            CAM::SET_CAM_ACTIVE(zoomCamera, true);
+            UpdateCamera();
+        }
+        if (n == 1)
+        {
+            camZoom -= camZoomRate;
+            CAM::SET_CAM_ACTIVE(zoomCamera, true);
+            UpdateCamera();
```
Last0xygen, line 51: „SET_CAM_ACTIVE“ and „UpdateCamera“ can be outside of the ifs, as they get executed either way 
Also they need to be executed every tick not every fifth or it would result in stuttering

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
```diff
@@ -0,0 +1,72 @@
...
+    zoomCamera = 0;
+}
+
+static RegisterEffect registerEffect(EFFECT_PLAYER_ZOOMZOOM_CAM, OnStart, OnStop, OnTick, EffectInfo
+    {
+        .Name = "Zoom Zoom Cam",
+        .Id = "player_zoomzoom_cam",
+        .IsTimed = true,
+        .IsShortDuration = true,
+        .IncompatibleWith = { EFFECT_FLIP_CAMERA, EFFECT_PLAYER_GTA_2, EFFECT_PLAYER_QUAKE_FOV, EFFECT_PLAYER_BINOCULARS }
```
Last0xygen, line 70: this effect has to be added to those other effects too :)

Last0xygen (changes requested):
Looking good so far :)

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
TheRealAyCe: Really? Cant't you just check both? Weird rule imo

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
Mo-xii: Not quite sur why it's needed, but done anyway!

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
Mo-xii: I think it's okay now. When is a namespace prefix needed?

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
Mo-xii: That's so much better! Thank you for the suggestion. It's way cleaner now, 1 if instead of 4

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
Mo-xii: Done

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
Mo-xii: Thank you for that. I did notice a bit of a stutter. I was going to have a look at it to solve it. You did it for me thank you! It's making a lot more sense to write it outside of the ifs.

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
Mo-xii: I've tested ingame and it's working fine without it I believe. If I start Binoculars then Zoom Zoom, or the other way around it always disable the first one before doing the second.
Of course I can still add the exceptions to the others effects if you want me to, but why are they required? I don't know what I'm missing.

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
Last0xygen: Namespaces are needed if they are not included which they are via the „stdafx.h“ Header

ChaosMod/Effects/db/Player/PlayerZoomZoomCam.cpp
ProfessorBiddle: I think it's meant more for code cleanliness than function. Maybe a list of effect groups that are incompatible with each other in a separate file might be good eventually

Mo-xii: Okay I understand. I've added it to the other effects. I also added Binoculars to Flip, GTA2 and Quake because it was missing. I hope it's okay

Issue #1056, opened 2021-04-14
[Effect Suggestion] Alt Strat


Da-Best:
**Name of effect**
Alt Strat

**Duration if timed effect (long or short timed?)**
Short

**Effect Description**
Holds down the character swap button or makes it look like the character swap button is held down.

anto4440:
This has been suggested before

Aljotzi:
> This has been suggested before

Do you have an issue cuz I couldn't find one when I looked through it

anto4440:
> > This has been suggested before
> 
> Do you have an issue as an example cuz I couldn't find one when I looked through it

Thats a bit harsh.. The one goddamn time i dont write a clarifier saying im not one hundred percent certain some guy comes and says i have problems.. My luck is not with me today

ProfessorBiddle:
> Thats a bit harsh.. The one goddamn time i dont write a clarifier saying im not one hundred percent certain some guy comes and says i have problems.. My luck is not with me today

i don't think that's what he meant.


this was suggested a bunch of times on the DarkViper discord before the github system was implemented

Aljotzi:
> > > This has been suggested before
> > 
> > 
> > Do you have an issue as an example cuz I couldn't find one when I looked through it
> 
> Thats a bit harsh.. The one goddamn time i dont write a clarifier saying im not one hundred percent certain some guy comes and says i have problems.. My luck is not with me today

Also I never claimed that you have problems. I was just wondering if there was an issue number with the suggestion. I completely forgot that people used Matt's server to submit stuff so that was my mistake.

Da-Best:
If it was on darkviper's discord then lets say i put it onto the website but here we can actually say our opinions on each topic unlike on viper's discord

Last0xygen:
Let’s all chill out...
It’s a valid request, there is no Chaos effect request on Matts discord anymore.
This is the place to request effects, nothing done wrong here.
I also see no other mention of this effect so it can stay until someone says otherwise :)

ubscal:
Added this https://github.com/gta-chaos-mod/ChaosModV/pull/1240

PR #1240, opened 2021-04-30
Added Effect: "Alt Strat"

ubscal:
Requested in https://github.com/gta-chaos-mod/ChaosModV/issues/1056#issue-857994207


ProfessorBiddle:
A different name for this might be a good idea for the people who don't watch Matt's twitch/youtube. Maybe 'Player Swap' can make a return?

ubscal:
> A different name for this might be a good idea for the people who don't watch Matt's twitch/youtube. Maybe 'Player Swap' can make a return?

Player Swap actually triggered the Player Swap Camera, This holds down the "Alt" key, which brings the menu up. I'm pretty sure its just called "Alt Strat" in GTAV Speedrunning, but I could be wrong.

Da-Best:
isnt player swap already an effect/ was an effect where it put you in the clouds 

I don't want people to be confused on alt strat and player swap

ubscal:
> isnt player swap already an effect/ was an effect where it put you in the clouds
> 
> I don't want people to be confused on alt strat and player swap

Player swap seems to have been removed in the new patch, this effect just holds down the left alt key. I'm working on redoing this effect to not just depend on the player swap key being on left alt.

ProfessorBiddle:
>  I'm working on redoing this effect to not just depend on the player swap key being on left alt.

the way it is now should be fine, 'INPUT_CHARACTER_WHEEL' should be the same regardless of control settings

https://github.com/MAFINS/MenyooSP/blob/master/Solution/source/Scripting/enums.h

ubscal:
> > I'm working on redoing this effect to not just depend on the player swap key being on left alt.
> 
> the way it is now should be fine, 'INPUT_CHARACTER_WHEEL' should be the same regardless of control settings
> 
> https://github.com/MAFINS/MenyooSP/blob/master/Solution/source/Scripting/enums.h

Ah thank you, I must've confused myself into thinking that my code only held down the left alt key.

Last0xygen:
"Quick Update" 😅 
Just tried the effect, could be avoided by bringing up the phone. 
But in general, while it was a funny idea its just not too different from "0.2x Gamespeed", so i don't think this would be a good addition to the mod. sorry

ubscal:
> "Quick Update" 😅 Just tried the effect, could be avoided by bringing up the phone. But in general, while it was a funny idea its just not too different from "0.2x Gamespeed", so i don't think this would be a good addition to the mod. sorry

imma be real with you i completely forgot I made this LMAO, and yea i totally agree with you.

Issue #1209, opened 2021-04-28
[Effect Suggestion] Melee Only


LordRygar:
**Name of effect**
Melee Only

**Duration if timed effect (long or short timed?)**
Long

**Effect Description**
Removes all Weapons except Melee Weapons from everyone for a persistently long time.

ubscal:
Added this https://github.com/gta-chaos-mod/ChaosModV/pull/1244

PR #1244, opened 2021-04-30
Added Effect  "Meele Only"

ubscal:
Based off of https://github.com/gta-chaos-mod/ChaosModV/issues/1209

Forces every Ped and the player to use their fists, they cannot select any weapons until the effect is over.


LordRygar:
Nice. It's Brawling Time!  Fists will be perfectly adequate. Might be OP to allow melee weapons.

Last0xygen:
The only difference from "Remove Weapons From Everyone"  is that you keep your weapons 🤔  i dont know if its that different :/

ubscal:
> The only difference from "Remove Weapons From Everyone" is that you keep your weapons 🤔 i dont know if its that different :/

True,

ubscal:
Sorry for accidently closing, fat fingered the button 🙃

Last0xygen:
Sooooo what to do next? As I don’t see a need for another effect which removes all weapons forever/a period of time, do we still need this effect? :)

ubscal:
> Sooooo what to do next? As I don’t see a need for another effect which removes all weapons forever/a period of time, do we still need this effect? :)

I think closing this would be a good option, Remove all weapons is too similar anyway.

PR #1283, opened 2021-05-04
Added Effect "Rocket Man"

Mo-xii:
With the potential introduction of "Fling Player" by Last0xygen, "Launch Player Up" might not be as appealing as before. So this is my attempt to have kind of a launch player up, but a bit more unique.

I would like to launch the player higher, but increasing the velocity while ragdolling doesn't seem to work.

I'm waiting for your suggestions and improvements. 
Thank you.


pongo1231:
Looks good, only thing that's completely redundant is the `DOES_ENTITY_EXIST` check for the player ped (which will always be the case so no need to check for it). I fixed it along with compilation issues caused by upstream changes.

PR #1294, opened 2021-05-06
Added Effect "Disable Random Direction"

Mo-xii:
Based on a effect from GTA San Andreas Chaos Mod by Lordmau5.

I think this effect could be a good alternative to "No FW/BW/L/R Movement". Whenever you're driving, you basically have to wait the end of the effect to do anything. By disabling only one direction, progress is still possible albeit more difficult.

For a reason I don't understand I couldn't only disable INPUT_MOVE_UP_ONLY for instance. So the workaround I found was to disable INPUT_MOVE_UD whenever UP is pressed.

As you can see there are a lot of ifs, so this is probably not a good way to do it. 

I'm welcoming any critics, suggestions and improvements.
Thank you


Numi43:
The point of the mod is not to disable the player´s chance to progress though. Thus bad suggestion.

Mo-xii:
Wait, what? Maybe I'm not really clear about what the effect is doing.

I'm actually giving the player more control over the game than the 2 previous effects. "Disable forewards / backwards movement" disable 2 directions and you can't drive anymore. I only disable one and thus you can still drive (maybe you'll have to drive backward, but you can still progress)

I'm confused. My goal is to do the exact opposite of what you're accusing me of.

Numi43:
Ah I did not get it correctly. Forgive me. Could be rather a meta effect tbh.

And could be named something like "Not This Way (L/R/U/D)"

ChaosMod/Effects/db/Player/PlayerNoMovRandom.cpp
```diff
@@ -0,0 +1,121 @@
...
+static void OnStart()
+{
+	int_rand_key = g_random.GetRandomInt(0, 3);
+}
+
+static void OnTick()
+{
+	if (int_rand_key == 0) // Disable Up
+	{
+		if (IS_CONTROL_PRESSED(0, 32) == true)
```
Last0xygen, line 18: You could just check with „ if (IS_CONTROL_PRESSED(0, 32))“ (compares to true)
But do you even need those checks? Can’t you just disable them even if they don’t press them? 🤔

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerNoMovRandom.cpp
Mo-xii: First thing I tried was : DISABLE_CONTROL_ACTION(0, 32, true)
So basically disable the INPUT_MOVE_UP_ONLY, but that doesn't work. The only way I found to fully disable moving forward was to disable INPUT_MOVE_UD which disable Up and Down at the same time, thus the workaround.

This is a mainly a problem when you walk. No problem when driving because the game is just using INPUT_VEH_ACCELERATE and INPUT_VEH_BRAKE. There is no "INPUT_VEH_AB" that would be used for accelerating and braking. I hope I make sense.

So in conclusion what I found so far : disabling INPUT_MOVE_UP_ONLY doesn't do anything if INPUT_MOVE_UD is still enable.

Thank you for the tip regarding the if!

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerNoMovRandom.cpp
```diff
@@ -0,0 +1,120 @@
...
+			DISABLE_CONTROL_ACTION(0, 149, true);
+		}
+		DISABLE_CONTROL_ACTION(0, 72, true);
+		DISABLE_CONTROL_ACTION(0, 88, true);
+		DISABLE_CONTROL_ACTION(0, 130, true);
+		DISABLE_CONTROL_ACTION(0, 138, true);
+		DISABLE_CONTROL_ACTION(0, 139, true);
+		DISABLE_CONTROL_ACTION(0, 281, true);
+	}
+	if (int_rand_key == 2) // Disable Left
```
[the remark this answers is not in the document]
Mo-xii: I think I've done it properly. I used "else if" and an "else" at the end.

pongo1231:
I'm surprised you got it to work since in my testing it only took effect if the opposite direction was disabled too (thus the "Disable Forward/Backward Movement" and "Disable Left/Right Movement" effects). Those can be replaced by this effect now.

PR #1350, opened 2021-05-12
Added Effect 'News Team'

Last0xygen:
Follows the player with a news helicopter


RosyMiranto18:
Is this News Team will broadcast news like in the Marathon and Legal Trouble, and is the camera moved to be from the Chopper's perspective?

Last0xygen:
https://streamable.com/wwew8t

Issue #1158, opened 2021-04-24
[Effect Suggestion] Spawn Flyable Broomstick


DEATHCON1:
**Name of effect**
Spawn Flying Broomstick

**Effect Description**
Spawns a flyable broomstick that controls the same way as the Oppressor MkII or the Flyable Broomstick Mod for GTA San Andreas.

pipecek:
Not possible.

PR #1365, opened 2021-05-14
Added Effect 'You're A Wizard, Franklin'

Last0xygen:
Spawns an invisible MK2 with an attached broomstick taht the player can drive
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/1158


Mo-xii:
I'm curious to know why you set the broom as not longer needed : "SET_OBJECT_AS_NO_LONGER_NEEDED(&broom)"

Because very quickly we just have a completely invisble Opressor laying around. And if you want to despawn it, shouldn't you despawn the Opressor aswell?

Last0xygen:
I tested it and it didn’t despawn for everything I tried 🤷‍♂️  did you manage to make it just despawn the broom? 🤔

Mo-xii:
Yeah tested this afternoon. If I walk 10m away and look away it despawns the broom everytime. After that I can just drive the invisible Opressor. 

That behavior is consistent with objects marked as no longer needed, like the ramp of the Ramp Jam effect.

Last0xygen:
Well f me then I guess. I’ll adjust it, thx for testing

Issue #680, opened 2021-02-27
[Effect Suggestion] there aren't cougars in this mission!


0rangeGhost:
**Name of effect**
there aren't cougars in this mission!

**Effect Description**
spawn a cougar/any hostile animal behind the player that will hunt player

Numi43:
A better name would be "Matto´s Nightmare" or "Speedrunner´s Nightmare".

SeanMirrsen:
Better yet, "Here There Be Cougars", referencing Bradbury's "Here There Be Tygers".

The effect could periodically cast rays (or fire invisible bullets) behind the camera, spawning a cougar wherever the ray/bullet lands. The key is to spawn the cougar behind the camera, not the player, so that the player never sees the cougar spawn.

TheRealAyCe:
There aren't _enough_ cougars in this mission!

Hackermanxdd:
THIS BETTER BE ADDED IN THE NEXT UPDATE

ProfessorBiddle:
> THIS BETTER BE ADDED IN THE NEXT UPDATE
>

It definetely will (not) if you keep asking like this

>

TheRealAyCe:
People work on what they want to work on. If none of the programmers want to code this, it will not be done. The suggestion phase is just to reject bad/duplicate ideas.

Da-Best:
The only thing that is the issue is that cougars only attack when you are on foot and not when you are in cars so if this happens when you are in a car then it basically does nothing.

SeanMirrsen:
Then for once it'll be an effect that only works if you're on foot, rather then only when you're in a vehicle.

pongo1231:
Part of the mod now

PR #1380, opened 2021-05-18
Added Effect "Hot Cougars In Your Area"

Mo-xii:
Similar to "Killer Clowns" but with cougars on fire.

#680


bulva2:
There aren't cougars in missions!

PR #1539, opened 2021-06-14
Ensure wanted levels persist through fake TP

kolyaventuri:
- Modifies fake teleport such that if you have a wanted level, it doesn't go away
- Ensures that if you get fake teleported into Fort Zancudo or LS Airport, it gives you the proper 4 or 3 star wanted level respectively
- Ensures that if you have a 4 or 5 star wanted level, while being teleported to LS Airport or Fort Zancudo respectively, that it is not lowered to the 3 or 4 star wanted level


ChaosMod/Effects/db/Player/PlayerTpController.cpp
```diff
@@ -291,10 +303,19 @@ static void OnStartFakeTp()
...
+	if (wanted == 0 || wanted < currentWanted)
+	{
+		wanted = currentWanted;
+	}
+
 	SET_PLAYER_WANTED_LEVEL(player, 0, false);
 	SET_PLAYER_WANTED_LEVEL_NOW(player, false);
 	SET_MAX_WANTED_LEVEL(0);
 
+	SET_FAKE_WANTED_LEVEL(wanted);
```
kolyaventuri, line 317: Noticing in testing that the fake wanted level has a different effect than the real one... No police are dispatched so the effect might not be as convincing this way. Could adjust the real wanted level alternatively

SeanMirrsen: I've already commented on the original commit that added the new script-pause protections, but pongo must've missed it.

With the script-pause being a thing, it should no longer be possible for the mission to fail due to temporary real wanted level. So the real wanted level should never be set to 0 to begin with, it should just be re-set to whatever it was before teleporting.

kolyaventuri: Good point! Could just ensure that any existing fake or real wanted level is persisted after the effect. Just set a real wanted level, and reset to original values after the fake TP is revealed

Last0xygen: The wanted level has to be set before the script-pause though, otherwise it wont be shown -> would potentially cause mission fails. this is "good enough" imo :)

Issue #1547, opened 2021-06-14
[Bug] Catto guns may cause crash


TheRealAyCe:
https://www.twitch.tv/videos/1055780849
~1:57:30

PR #1550, opened 2021-06-14
Fix (maybe) for catto guns

Last0xygen:
Sets a max amount of cats that can be spawned
Fix https://github.com/gta-chaos-mod/ChaosModV/issues/1547

### DIFF
--- a/ChaosMod/Effects/db/Peds/PedsCatGuns.cpp
+++ b/ChaosMod/Effects/db/Peds/PedsCatGuns.cpp
@@ -1,5 +1,8 @@
 #include <stdafx.h>
 
+static int MAX_CATS_COUNT = 20;
+static std::vector<Ped> currentCats;
+
 static void OnTick()
 {
 	static const Hash catHash = GET_HASH_KEY("a_c_cat_01");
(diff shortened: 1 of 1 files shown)


ChaosMod/Effects/db/Peds/PedsCatGuns.cpp
```diff
@@ -1,5 +1,8 @@
 #include <stdafx.h>
 
+static int MAX_CATS_COUNT = 20;
```
kolyaventuri, line 3: Is it worth making this configurable? Even if that's a "hidden" config?

Last0xygen: what do you mean "configurable"?
defining it on top? letting the user change it? 🤔

kolyaventuri: More or less? At least allow for more cats on more powerful machines? That can always be a later improvement though

Last0xygen: I don’t think it’s depending on „powerful machines“ but on the ped limit in game. As Long as we don’t have that implemented we should not be throwing objects around imho

pongo1231:
Was there any specific reason we didn't merge this yet 😄? Needs to be updated though.

PR #1559, opened 2021-06-16
Fix Teleport Effects

Last0xygen:
* Fake Teleport
  * Teleports the player not to different places if in a vehicle or not -> adjusted it to repllicate the original teleport
* Teleport to Store
  * Adjusted Positions for Ammunitions as you can get stuck behind the counter

Also i apparently updated minhook 🤔  😂


kolyaventuri (approved):
Nice!

PR #1588, opened 2021-06-18
Added Effect: "Spawn Bedmobile"

Mo-xii:
Spawn a "road legal" bed.


Mo-xii:
Just realised to bed is an online prop.

PR #1602, opened 2021-06-21
Added Effect: "Monster Head"

Mo-xii:
Spawn a monster head attached to the player

https://i.imgur.com/BHVn2HH.png

It's not perfect, as you can see, the monster's neck is showing through the player's chest.

Any idea on how to improve it?


ChaosMod/Effects/db/Player/PlayerMonsterHead.cpp
```diff
@@ -0,0 +1,27 @@
...
+
+static void OnStart()
+{
+	LoadModel(monsterHash);
+
+	Ped player = PLAYER_PED_ID();
+	int bone = GET_PED_BONE_INDEX(player, 0x60F2);
+	Vector3 playerPos = GET_ENTITY_COORDS(player, false);
+
+	Object monsterHead = CREATE_OBJECT(monsterHash, playerPos.x, playerPos.y + 2, playerPos.z, true, false, false);
```
Last0xygen, line 17: Create it using "CreatePoolProp" instead -> no "SET_ENTITY_AS_NO_LONGER_NEEDED" needed then

Last0xygen (changes requested):

PR #1612, opened 2021-06-22
Added effect "Look At This!"

Goorakh:
Rotates the camera to face a random ped


TheRealAyCe:
No visibility check?

Goorakh:
> 
> 
> No visibility check?

You mean having a check to see if the ped is visible to the camera?

TheRealAyCe:
Yes, or did you test it and didn't encounter any cases where the camera would focus a building wall because there was a ped behind it? Maybe that's even good. Just an idea.

Goorakh:
> 
> 
> Yes, or did you test it and didn't encounter any cases where the camera would focus a building wall because there was a ped behind it? Maybe that's even good. Just an idea.

I thought about adding a check like that, but I think it would just get blocked by small objects like vehicles or poles, the max range should deal with the issue of it focusing on a ped through a wall though. I could probably make it refocus if the target ped hasn't been visible for X seconds, but I just didn't think that was necessary, it was never an issue in my testing at least.

SeanMirrsen:
Personal opinion: I'd have called the effect "Would you look at that!". I think having a more expressive turn of phrase as the effect name is funnier. :)  
Alternatively, "Look at that guy", or "Who is that guy" or something, since you're making the camera look at a ped rather than an object.

Also, there are plenty of instances where the player is far away from basically any peds at all, for instance when flying, or driving through various empty mountains and deserts. Having the effect point at something else, even if it's a random point on the ground, would be nice. Or just spawn a random ped and focus on them if there aren't any around.

TheRealAyCe:
Yes, some fallback targets would be nice.

Goorakh:
> 
> 
> Personal opinion: I'd have called the effect "Would you look at that!". I think having a more expressive turn of phrase as the effect name is funnier. :)
> Alternatively, "Look at that guy", or "Who is that guy" or something, since you're making the camera look at a ped rather than an object.
> 
> Also, there are plenty of instances where the player is far away from basically any peds at all, for instance when flying, or driving through various empty mountains and deserts. Having the effect point at something else, even if it's a random point on the ground, would be nice. Or just spawn a random ped and focus on them if there aren't any around.

Yeah, I guess a fallback couldn't hurt, I'll add that, thanks for the feedback guys :)

TheRealAyCe:
Animals count as Peds too, right? So in a forest your camera may focus some animal, which sounds good too :D

Goorakh:
> 
> 
> Animals count as Peds too, right? So in a forest your camera may focus some animal, which sounds good too :D

Yes, any living ped that is not the player can be focused.

Goorakh:
Closing this partially because I don’t think it’s a very good effect after playing with it for a bit, but mostly because the code is quite bad and buggy, and I don’t see any point in rewriting it.

PR #1642, opened 2021-06-27
Added Effect "Spawn Quarreling Couple"

juhana:
Spawns Debra with a pistol and Floyd with a knife.


RosyMiranto18:
I think this idea works. However, they will attack each other but if you try to interfere in any way, you will be attacked too.

PR #1645, opened 2021-06-27
Added Effect "Vehicle Swap"

juhana:
All peds try to get into the closest vehicle to them that's not their current vehicle.


SeanMirrsen:
Suggested alternate name: "Chaange places!"

May be forgotten nowadays, but it's a running gag with the mad hatter robot in Futurama.
As seen in:
https://www.youtube.com/watch?v=8tYXfssLOSM
https://www.youtube.com/watch?v=chjVoVcVi3A

Issue #340, opened 2021-01-11
Key bind effects.


MaatRolo:
Is your feature request related to a problem? NO

**Describe the solution you'd like**
Key bindings - I'm thinking 'Ctrl+(NUM1-NUM9)' - to start/stop effects

**Describe alternatives you've considered**
None Alternatives

**Additional context**
I'd like to be able to hot key specific effects to trigger them on command.   Most useful example I can think of is Suicide.  Having it bound to CTL+NUM1 if I'm stuck in because of an effect but don't wish to restart mod or page through the (CTRL+,) menu.  This could also be used to bind snow to a Key combo to change weather if you wanted.  

Let me know if this needs further explanation

MoDerPo:
There are 270 effects.

MoDerPo:
use mod menus if you want to start a specific effect not the chaosmod

SeanMirrsen:
I think the request is to be able to "vote" 1/2/3/4 like Twitch chat would, on random upcoming effects, without actually having a Twitch chat or being online at all. With gamepad buttons or keyboard shortcuts.

MoDerPo:
oh yeah I had that idea too (#535)

PR #1757, opened 2021-07-09
Added Feature 'Shortcuts'

Last0xygen:
Adds the ability to add 'Shortcuts' to active effects via the config app so a player can either manually trigger it by keypress or even set up a bot to trigger a key press for donations, gifts or channelpoints. :)
Added the F-Keys and Numpad-keys as possible settings.
Also changed the way effects get saved to allow commas in effect-name overrides

Fix https://github.com/gta-chaos-mod/ChaosModV/issues/340


TheRealAyCe:
Now Matto can charge viewers 5 AUD to spawn Griefer Jesus! ^^

Issue #1911, opened 2021-07-23
[Effect Suggestion] Spawn Angry Chimp


CHETEMI59:
**Name of effect**
Spawn Angry Chimp

**Effect Description**
Spawns an enemy chimp with a stone hatchet.

Last0xygen:
Fixed https://github.com/gta-chaos-mod/ChaosModV/pull/1915

Issue #1407, opened 2021-05-21
[Effect Suggestion] Spawn Piglin Brute


RosyMiranto18:
**Spawn Piglin Brute**

**Duration if timed effect (long or short timed?)**
One-time

**Effect Description**
Summons a hostile NPC Pig/Boar carrying either a Battle Axe (but he’s very tanky), or a Stone Hatchet with the hatchet’s special ability permanently activated.

Last0xygen:
Fixed https://github.com/gta-chaos-mod/ChaosModV/pull/1915

PR #1915, opened 2021-07-24
Added Effect "Spawn Angry Chimp"

Marvisak:
Spawns chimp ped with stone hatchet which attacks player and every other ped


RosyMiranto18:
Isn't this basically the same idea as my Piglin Brute one with a different animal? At #1407

Marvisak:
> Isn't this basically the same idea as my Piglin Brute one with a different animal? At #1407

I saw this as effect suggestion exactly as it is, #1911

PR #1977, opened 2021-07-31
Update PlayerSuperRunJump.cpp

zalanveszelovszki:


TheRealAyCe:
So this just makes the speed x4 instead of ~x1.5? No explanation?

PR #1985, opened 2021-08-01
Added effect "Vehicles Disappear On Impact"

juhana:
Despawns vehicles when they hit something.


Last0xygen:
„Remove Current Vehicle“ already is broken enough for mission entities imho 😅

TheRealAyCe:
Can the effect be made to appear with reduced likelihood, or group together with "Remove Current Vehicle"? Would be a shame to throw the code away, so to speak.

SeanMirrsen:
I have an alternate suggestion.

As in I literally made an alternate suggestion: #1986  :P 

I think it could be more interesting to have an effect where *anything* suddenly disappears rather than dying/exploding, and then have this effect happen as a combination with that one, rather than be a standalone thing that is a bit too similar to the existing Vehicles Explode On Impact effect.

PR #2145, opened 2021-08-18
Added effect "Make Random Stunt Jump"

Reguas:
Teleports the player to one of stunt jump locations, gives him the right angle and speed to complete the jump. If the player is on foot, he is given a Pegassi Bati 801 motorbike.

Warning: some of the jumps are not consistent, and some can be completed only with a bike, but it makes an effect more chaotic, I guess.


Reguas:
Reopening this to avoid force-pushing

PR #2183, opened 2021-08-22
Updated Blimp Strat

Reguas:
Now the "Daveyy" cutscene triggers as player falls down. The cutscene will always contain Michael regardless of the current character, although when it ends, the player gains back control of a current character.

Also slightly changed the position of the blimp in order to make the fall more persistent.

### DIFF
--- a/ChaosMod/Effects/db/Player/PlayerBlimpStrats.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerBlimpStrats.cpp
@@ -10,6 +10,7 @@ static void OnStart()
 	bool cutscenePlaying = IS_CUTSCENE_PLAYING();
 
 	Hash blimpHash = GET_HASH_KEY("blimp");
+	Hash daveHash = GET_HASH_KEY("ig_davenorton");
 
 	LoadModel(blimpHash);
 	
@@ -27,6 +28,8 @@ static void OnStart()
 	SET_VEHICLE_FORWARD_SPEED(veh, 45);
 	TASK_LEAVE_VEHICLE(player, veh, 4160);
 
+	
+
 	int waited = 0;
 
 	while (!IS_PED_GETTING_UP(player) && waited < 100)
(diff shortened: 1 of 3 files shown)


Last0xygen:
Why block the Player for an additional 10 seconds? 🤔

Reguas:
Yeah, could make it skippable, but that would cause problems while in missions, showing "mission failed" screen while playing the cutscene. Plus, 13 seconds for an effect isn't _that_ much. Gravity Field and Autopilot, for example, last twice as long, and player also can do nothing.

rollschuh2282:
there is a way to change a character during a cutscene with this native.
https://alloc8or.re/gta5/nativedb/?n=0xE40C1C56DF95C2E8
maybe a check could get implemented so the current player appears in the cutscene rather than always michael.

Reguas:
Thank you, successfully changed this.

TheRealAyCe (commented):
What are the hashes for? Shouldn't they be defined globally somewhere instead of copypasted into specific effects?

Reguas:
Idk what were they doing there 😐

ChaosMod/Effects/db/Player/PlayerBlimpStrats.cpp
```diff
@@ -1,24 +1,68 @@
...
+
+		REGISTER_ENTITY_FOR_CUTSCENE(player, "MICHAEL", 0, 0, 64);
+		
+		START_CUTSCENE(0);
+		WAIT(8500);
+		STOP_CUTSCENE_IMMEDIATELY();
+
+		REMOVE_CUTSCENE();
+
+		Ped pedDave = CreatePoolPed(4, daveHash, -442.2f, 1059.25f, 326.66f, 180.6f);
```
pongo1231, line 61: Would be better to just spawn a normal ped and use `SET_PED_AS_NO_LONGER_NEEDED` instead, no need to make Dave persistent

Reguas: Gotcha

rollschuh2282:
not sure if needed, but maybe a check for the real mission would be smart.
bc doesn´t sound very good if the game tries to start the mission and then the script loads the cutscene as well.
sounds like it could break the mission script.

Reguas:
> not sure if needed, but maybe a check for the real mission would be smart

No, that's fine. `IS_CUTSCENE_PLAYING()` is enough for the effect to work correctly

OnlyRealNubs:
Maybe we should set the player invulnerable, in the possibly inevitable case that the player could die when they are jumping out.

Reguas:
> If the player is already at low HP, and they hit the ground when it is deactivated, it could kill them.
Wdym, it's active unless the player is on the ground

OnlyRealNubs:
When I tested, I would get damage from hitting the pavement.

OnlyRealNubs:
> When I tested, I would get damage from hitting the pavement.

@Reguas After doing more testing. This only happens my fps is locked to 30, from the `Console Experience` effect.

Reguas:
Kinda sus, but ok, moved invincibility deactivation after the cutscene

OnlyRealNubs (approved):

PR #2236, opened 2021-08-30
Allow More Concurrent Effects

Last0xygen:
Allows more effects to be run simultaniously and not be removed if 8 effects are active.
Can be easily observed with "Combo-Time": It does currently stop timed effects if "too many" effects get started
* Allow more timed effects to run 
* Still clean other effects as usual
* dynamically reduce spacing between the active effects to always use all the space


ChaosMod/Components/EffectDispatcher.cpp
```diff
@@ -125,15 +127,11 @@ void EffectDispatcher::UpdateEffects()
...
 
 			if ((effect.m_fMaxTime > 0 && effect.m_fTimer <= 0)
-				|| ((!effectData.IsMeta)
-					&& (effect.m_fTimer < -m_usEffectTimedDur + (activeEffectsSize > 3
-						? ((activeEffectsSize - 3) * 20 < 160
-							? (activeEffectsSize - 3) * 20
-							: 160)
-						: 0))))
+				|| (!effectData.IsMeta && (activeEffectsSize > maxEffects  || (effect.m_fMaxTime > 0 && ShouldRemoveEffectForTimeOut(effect.m_fTimer, activeEffectsSize, effectCountToCheckCleaning))))
+				)
```
kolyaventuri, line 131: It's probably worth splitting this into a few variables, since this this statement is a bit difficult to parse

kolyaventuri: But it's definitely better than the nested ternaries

kolyaventuri (approved):
LGTM. Worth making it configurable? Thinking of lower-end systems where combo time might cause issues. That being said, the solution there might be "disable combo time"

ProfessorBiddle:
> Worth making it configurable?

I think that would be a good idea, even excluding low end pc's some high end machines can handle a lot more than others

MoneyWasted:
I would make it an option that can be changed with the config app rather than hard coding it.

Last0xygen:
> I would make it an option that can be changed with the config app rather than hard coding it.

well its not "hardcoded" so to speak, but depending on the available space for the effects to display.
But yeah, i added an option in the config app to modify it

ChaosMod/Components/EffectDispatcher.cpp
Last0xygen: done

kolyaventuri (approved):

pongo1231:
Looks good! 👍

Issue #1571, opened 2021-06-17
[Effect Suggestions] Legitimate businessman


NIGHTFURY2111:
Legitimate businessman:- 
• Remove current vehicle (if any),
• Remove all weapons ( or simply force to use only fists),
• Spawn Simon ( for a fist fight)

Extra:- 
• Simon can have extra health 
•Simon can say his lines from cutscenes

PR #2254, opened 2021-09-02
Added effect "Legitimate Businessman"

Reguas:
Suggested in #1571.
Removes all weapons and money from the player, spawns hostile Simeon for a fist fight. If the player is in vehicle, either Lamar or Franklin spawns and hijacks it.


RosyMiranto18:
I think we should remove the losing money part of this effect. Because that would mean that we would have another Poor Boy effect and it would be too cruel.

Reguas:
Seems like money isn't a big issue in Chaos. We have at least 3 effects that give money (HESOYAM, Midas Touch and Money Rain), and only one that takes it, so it's kinda disbalanced. Though that's a funny addition to Simeon: he takes all your money, weapons and vehicle.

RosyMiranto18:
Well, your have a point about that, but when the effect that takes away your money takes ALL of your money, on all characters, and their Shares as well, i think one alone is good enough. That is unless if Poor Boy is nerfed and only takes away the money that your current character have.

Issue #741, opened 2021-03-06
[Feature Suggestion] Adjustable effect scaling


SeanMirrsen:
This is a simple idea that is likely a pain to implement, like most simple ideas.

The feature would call for, somewhere in the function that calls up an effect, an extra variable to be passed along. That variable, is then tied as a multiplier into some key values that allow the effect to be _scaled in magnitude_, by passing it a simple multiplier variable. The variable is then exposed to the user in the settings app, as yet another thing they can tweak about any given effect.

The practical upshot of this change, is that effects become adjustable in regards to more than just their length. I'll describe some examples I can think of:

- Gravity Field or Forcefield: change effect radius, or push/pull strength. 
- Earthquake, Doomsday: change magnitude of force affecting objects.
- Spawn <object/vehicle/enemy>: spawn one instance for every full point of magnitude, rounded up (so magnitude 0.8 still spawns one, magnitude 1.2 spawns two, 2.1 spawns three, etc)
- Gamespeed modifiers: Divide the multipliers by the magnitude (so 0.5x speed at magnitude 2 would result in 0.25x speed)
- Meteor rain/Whale rain/Vehicle rain/Killer clowns or other effects that spawn or activate something periodically: Divide the delay by the magnitude (so higher magnitude = faster)

Many more examples out there. However, the feature wouldn't work for a number of effects. Teleports would be unaffected, effects that remove weapons/vehicles would be unaffected, recolors and other randomizers would be unaffected. Likely a lot of others. So it won't be necessary to change every single effect to accommodate the feature.

This feature would allow users to fine-tune the effects to their liking, like reducing the strength of gravity field, or increasing the number of rocks spawned by meteor rain if they so wish.
Additionally, having the variable available, would allow metaeffects to alter it as well.

PR #2257, opened 2021-09-02
Added meta effect: 2x Chaos

Goorakh:
Multiplies the "effect" of every effect by 2, in general, when normally spawning 1 object, it now spawns 2, and timers are generally cut in half.

Now, of course this doesn't work for every effect, the vehicle color and weather changing effects for example, but during my testing, I rarely encountered a situation when the multiplier would have no effect, but yes, it is possible this meta effect doesn't have any effect, but I would argue the changes of it are pretty low.

I also included a debug button for switching between a multiplier of 1 and 2 (keybind is +), which makes it easier to test it without having to enable and disable the effect over and over.

Of course, if you intend for meta effects to always have an effect on the mod, sure, maybe this wouldn't be a good addition, but I personally think it's fine that it's not guaranteed to take effect.


kolyaventuri (changes requested):
There's gotta be a better way to do this, vs just adding extra code to every single effect... I like the idea, but this adds a serious maintenance and development burden.

Goorakh:
> 
> 
> There's gotta be a better way to do this, vs just adding extra code to every single effect... I like the idea, but this adds a serious maintenance and development burden.

Is it really a big burden though? I didn't spend that long on each effect making it multiplied (at least compared to the amount of time you might put into making the effect itself)

But yeah, the lazier approach would be to run OnStart and maybe OnTick twice for every time they normally would be called, but that just seemed really hacky to me.

kolyaventuri:
- Some effects on lower-mid end machines can cause crashes (think: earthquake or doomsday), having them be doubled would only increase the chance of a hard crash
- If a bug occurs, we now need to know if it was the regular version or the 2x version, and then work through hunting down what about the 2x version caused that
- Every new effect now needs a doubled version (when applicable)

It might not take a lot of time per effect, I see your point there, but there's the chance of it causing a number of issues / burden downstream. With the processing power part in particular, there's a non-zero chance that people would just disable the meta-effect all together. Which would be a bummer.

Goorakh:
> 
> 
>     * Some effects on lower-mid end machines can cause crashes (think: earthquake or doomsday), having them be doubled would only increase the chance of a hard crash
> 
>     * If a bug occurs, we now need to know if it was the regular version or the 2x version, and then work through hunting down what about the 2x version caused that
> 
>     * Every new effect now needs a doubled version (when applicable)
> 
> 
> It might not take a lot of time per effect, I see your point there, but there's the chance of it causing a number of issues / burden downstream. With the processing power part in particular, there's a non-zero chance that people would just disable the meta-effect all together. Which would be a bummer.

Hm, yeah, but I feel like if there was no multiplier to any potentially unstable effect, it would only be able to apply to maybe 10 effects or so. I do appreciate the feedback though, I didn't think of this at all when making it :)

kolyaventuri:
Totally! I recommend opening an issue before embarking on a 170 file monster of a PR 😛  Can iron out the kinks or figure out how to make it make sense before all the effort is put in

Goorakh:
> 
> 
> Totally! I recommend opening an issue before embarking on a 170 file monster of a PR 😛 Can iron out the kinks or figure out how to make it make sense before all the effort is put in

Honestly this was more of a personal challenge than anything, I had the idea, and wanted to see if I could implement it in a way I would be satisfied with, getting it approved would just be a nice bonus. But yes, next time I will open an issue for this kinda stuff :P

SeanMirrsen:
I made an issue with this suggestion before, #741 . My suggestion was basically working a *variable* into various effects, representing the "strength" of an effect. It wouldn't have to be 2x or any integer multiplier, it (hopefully) wouldn't need much extra code besides FOR loops around any spawning functions (and occasional math to add in the variable properly), it would just "magnify" specific parameters of most effects by the set variable.  

Looking at the code here, it seems like that's exactly what's being done. Except I would, if possible, expose the variable to settings, and make it per-effect. So that people could tone down (or turn up) the severity of Earthquake and Doomsday if they wanted, for example. And the metaeffect(s) would add a multiplier to that per-effect severity variable.

TheRealAyCe:
I think the changes made are not all that invasive and should not add a huge maintance burden, since essentially everything that was done is add a multiplier, necessitating the use of loops for many things. But that seems OK to me. One might also offer another way of registering effects that will just invoke the provided code multiple times, then not every single effect would need to know how to use the variable. Useful for "Spawn X" maybe.
I'm not a maintainer nor have I done much in C++, but the global variables and non-object nature of the effect code sadly seems to limit the ability to "cleanly" invoke effects twice at the same time.

I like it!

JustTunerYT:
I know this pull is 5 years old by now, but i've got my idea
Imo effects need to either get 'if's or Xs and that meta would either make an agreement on that 'if' or change the X between 0 and 3 but not 1. If that meta effect did have a range of numbers, in config there would be an option on what consistent number that could be.
It would require a lot of work in existing effects, but at least it would weigh less in the size of effect files (or to better explain it, no need to double code it)

PR #2281, opened 2021-09-05
Added effect 'Spinning Props'

Goorakh:
Like Spinning Peds but with props


TheRealAyCe:
Ah, it's actually correct now, no? So slow-motion will spin them slowly, and high framerate will still spin them just as fast as normal, not way faster.

PR #2290, opened 2021-09-07
Remove FPS-dependency from 'Spinning Cam'

ProfessorBiddle:
Spins like it would at ~60fps regardless of framerate/tickrate now.


pongo1231:
While this will ensure the camera won't rotate faster at a higher framerate, this will still rotate the camera at a slower pace in cases where the frametime is higher than 6 ms. A better solution would be to use `GET_FRAME_TIME` as a multiplier instead, which will ensure the camera will always rotate at the same rate.

PR #2291, opened 2021-09-07
Made random peds spawned also get random clothes

Goorakh:
Code copied from "Randomize Player Clothes"


juhana:
I tried this and sometimes it spawns peds with quite prominently exposed genitals which might cause problems to streamers.

Goorakh:
> 
> 
> I tried this and sometimes it spawns peds with quite prominently exposed genitals which might cause problems to streamers.

Oh, yeah that could definitely be an issue, didn't think of that.

pongo1231:
This should be fine currently. If any issues regarding that arise in the future - worst case, we'll manually maintain a blacklist of peds and variations which contain nudity.

PR #2345, opened 2021-09-13
[Feature Addition] Add "Login With Twitch" button

kolyaventuri:
Replaces the OAuth copy-paste flow with a proper `Login with Twitch.tv` button. 

Requires a twitch dev app to be set up, and the client ID + secret provided via a `.env` file in the `ChaosApp` folder.

image


ConfigApp/MainWindow.xaml.cs
```diff
@@ -386,6 +390,44 @@ void TwitchTabHandleAgreed()
...
+
+        void SetupTwitchLogin()
+        {
+            if (m_twitchAuth.LoggedIn)
+            {
+                twitch_logout_text.Content = "Logged in as " + m_twitchAuth.Username + " (Log Out)";
+
+                twitch_logout_text.Visibility = Visibility.Visible;
+                twitch_login_button.Visibility = Visibility.Hidden;
+                return;
```
Last0xygen, line 407: not needed

ConfigApp/MainWindow.xaml.cs
```diff
@@ -386,6 +390,44 @@ void TwitchTabHandleAgreed()
...
+            {
+                m_twitchAuth.Username = "";
+                m_twitchAuth.OAuthToken = "";
+                WriteTwitchFile();
+                ParseTwitchFile();
+
+                return;
+            }
+
+            m_twitchAuth.SpawnLogin();
```
Last0xygen, line 430: use "else" instead of return

Last0xygen (changes requested):

Last0xygen (approved):

Reguas:
Where do ypu get twitch client ID and secret & isn't it unsafe to store its raw values in the .env file?

Reguas:
(sorry, accidentally deleted previous message)
Where do you get twitch client ID & secret in release and isn't it unsafe to store its raw values in the .env file?

PR #2406, opened 2021-09-21
Fix "Invert Gravity"

Goorakh:
The Memory::SetGravityLevel() function was used as a multiplier, which resulted in very weak upwards gravity. Using -9.8 instead of -1 removes the need to use the APPLY_FORCE_TO_ENTITY_CENTER_OF_MASS native to make objects fly upwards properly, which fixes the "ass up" pose all peds used to have during Invert Gravity.

### DIFF
--- a/ChaosMod/Effects/db/Misc/MiscGravityController.cpp
+++ b/ChaosMod/Effects/db/Misc/MiscGravityController.cpp
@@ -69,21 +69,25 @@ static void OnStartInvert()
 
 static void OnTickInvert()
 {
-	Memory::SetGravityLevel(-1.f);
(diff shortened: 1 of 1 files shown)


TheRealAyCe:
Well, by now the ass-up is sort part of the fun and chaos. :)

Goorakh:
> 
> 
> Well, by now the ass-up is sort of part of the fun and chaos. :)

Fair enough haha, but it always bugged me that vehicles would get practically no lift at all, and everything went up at inconsistent speeds, it's just a little nit-pick of mine, and I thought I'd open a PR for it because why not

PR #2461, opened 2021-10-01
Smoother Zoom Zoom Cam movement

juhana:
Makes the Zoom Zoom Cam effect follow a sin curve instead of doing a hard bounce when it changes direction.


kolyaventuri (approved):
Solid!

PR #2519, opened 2021-10-15
Added effect: Beat the train

casperge:
Teleports player on the road in North Yankton. It spawns the player in a moving prologue getaway vehicle together with a companion brad.
Naamloos (2)


casperge:
Exactly why I closed it, it is just a bad idea

PR #2551, opened 2021-10-20
Added 20 extra possible display names for "Nothing"

casperge:
Added the following possible display names to "Nothing". The display name will change back to "Nothing" after 25 seconds. The effect will still be registered as "Nothing" and also show up like that in twitch voting.

"Nothing",
"All Peds Are Peds",
"Teleport To Current Location",
"Expanded & Enhanced",
"Spawn Air",
"Destroy All Destroyed Vehicles",
"Kill All Dead Peds",
"+0 Wanted Stars",
"Jesus Watches Over You",
"Set Time To Current Time",
"Set Player Into Current Vehicle",
"All Cops Are Cops",
"Aim to Point Gun",
"Give Everyone A Nose",
"Teleport Player A Few Millimeters",
"Flying Birds",
"Teleport All Elephants To Player",
"PC Experience",
"All Enemies Attack Player",
" " <- (actually nothing, will leave the player confused)

### DIFF
--- a/ChaosMod/Effects/db/Misc/MiscNothing.cpp
+++ b/ChaosMod/Effects/db/Misc/MiscNothing.cpp
@@ -1,6 +1,38 @@
 #include <stdafx.h>
+static std::string options[15] =
+{ "Nothing",
+"All Peds Are Peds",
+"Teleport To Current Location",
+"Expanded & Enhanced",
+"Spawn Air",
+"Destroy All Destroyed Vehicles",
+"Kill All Dead Peds",
+"+0 Wanted Stars",
+"Jesus Take No Wheel",
+"Set Time To Current Time",
+"Set Player Into Current Vehicle",
+"1x Vehicle Engine Speed",
+"You Aren't Famous",
+"GTA 5",
+"Crimes Are Illegal" };
 
-static RegisterEffect registerEffect(EFFECT_NOTHING, nullptr, nullptr, nullptr, EffectInfo
+
+static void OnStart()
+{
+
+	int start = GET_GAME_TIMER();
+	std::string override = options[g_Random.GetRandomInt(0, options->size() - 1)];
+	while (start + 25000 > GET_GAME_TIMER())
+	{
+		WAIT(0);
+		g_pEffectDispatcher->OverrideEffectName(EFFECT_NOTHING, override);
+
+
+
+
+	}
+}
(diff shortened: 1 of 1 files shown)


ChaosMod/Effects/db/Misc/MiscNothing.cpp
```diff
@@ -1,6 +1,40 @@
...
+" "};
 
-static RegisterEffect registerEffect(EFFECT_NOTHING, nullptr, nullptr, nullptr, EffectInfo
+
+static void OnStart()
+{
+
+	int start = GET_GAME_TIMER();
+	std::string override = options[g_Random.GetRandomInt(0, options->size() - 1)];
+	while (start + 25000 > GET_GAME_TIMER())
```
kolyaventuri, line 30: 25s might be too long, maybe like. 5 - 10?

ChaosMod/Effects/db/Misc/MiscNothing.cpp
```diff
@@ -1,6 +1,40 @@
...
+"All Enemies Attack Player",
+" "};
 
-static RegisterEffect registerEffect(EFFECT_NOTHING, nullptr, nullptr, nullptr, EffectInfo
+
+static void OnStart()
+{
+
+	int start = GET_GAME_TIMER();
+	std::string override = options[g_Random.GetRandomInt(0, options->size() - 1)];
```
Last0xygen, line 29: Please rename it so it’s not a keyname

ChaosMod/Effects/db/Misc/MiscNothing.cpp
Last0xygen: Why this loop? Wouldn’t it be „funnier“ if it just appears and  confuses and not be obviously „the nothing effect, that always does this“?

kolyaventuri: Take a look at how it's being used in the Fake Teleport OnStart -> https://github.com/gta-chaos-mod/ChaosModV/blob/master/ChaosMod/Effects/db/Player/PlayerTpController.cpp#L281

You should just be able to set it to what you want, and then wait and it'll change back after the delay. No need for a loop. 

```cpp
g_pEffectDispatcher->OverrideEffectName(...);
WAIT(25000);
```

PR #2629, opened 2021-11-28
Added effect: "Wiggle-Wiggle"

Reguas:
This effect is based on the GTA 3: Definitive Edition [bug](https://www.youtube.com/watch?v=jk6r5ZK-uQw): when you continuously turn a car left and right, its size increases. 
I believe there are more suitable names for an effect, so feel free to share your ideas.

Also I'm sure that memory manipulations should be moved to a separate class, I don't know where exactly though.


MoneyWasted:
Definitely worthy of being added to the mod. Also, I checked the pointers, and they seem to check out and are valid! :)

ChaosMod/Effects/db/Vehs/VehsWiggleChangesSize.cpp
```diff
@@ -0,0 +1,115 @@
...
+			}
+			if (length - savedLength < 0.1f)
+			{
+				MultiplyMatrix(getScriptHandleBaseAddress(vehicle), multiplier);
+			}
+		}
+	}
+}
+
+static RegisterEffect registerEffect(EFFECT_WIGGLE_CHANGES_SIZE, nullptr, nullptr, OnTick, EffectInfo
```
kolyaventuri, line 109: Needs an `OnStop` handler to reset to the original size

Reguas:
Not a great effect.

Mo-xii:
> Not a great effect.

I have to disagree. I love this effect and so does my chat. I increased the coefficient to give it some extra wiggling 👍

PR #2644, opened 2021-12-10
Change Memory::FindPattern Implementation

Parik27:
There are certain issues with the FindPattern implementation currently used in the Chaos Mod code-base that this pull request aims to resolve:

* FindPattern doesn't support patterns ending with trailing "?", and will cause a crash if they're used.
* Memory::Init needs to be called before FindPattern or the function will not return a result
* The implementation used fails to match certain patterns in the executable.

The pull-request changes the implementation used to one from CFX in effort to fix these issues and also to allow additional pattern related functions to be added (one used by DLC Despawn fix for example).


pongo1231:
Thanks!

PR #2645, opened 2021-12-10
Crash fixes for APPLY_FORCE_* natives

Parik27:
* Effects like Gravity Field and Doomsday tend to crash the game because of the physics engine's 228 (default) collider limit. Once the limit is reached, the game can no longer spawn any new dynamic entities and will crash as a result. 
* The PR adds a function with checks for these natives to ensure that enough colliders are available for the rest of the game.

Note: Requires #2644 to be merged because patterns used in the checks fail to match with the current FindPattern implementation.


pongo1231:
Cheers!

PR #2653, opened 2022-01-01
Add effect "Tiny Vehicles"

Goorakh:
Makes most vehicles half size (bikes doesn't work very well, but every other vehicle seemed to work).

The bulk of this effect is just copied from #2629 because I have absolutely no clue how to find memory offsets.


MoneyWasted:
Looks good and works for me all the points check out on the latest Steam version of the game. just calling on pogo to do the final checks. 😀

MoneyWasted (approved):

pongo1231:
7faefae77d9f211316279907a2ea50fde12532ce

PR #2655, opened 2022-01-05
Add effect 'Flipped HUD'

Goorakh:
Mirrors the game and mod HUD, almost all HUD elements are moved to the opposite side of the screen, effect timer goes from right to left, active effects are anchored to the left side of the screen.


pongo1231:
Nice work with the effect!

PR #2657, opened 2022-01-05
Add effect "Teleporter Malfunction"

Goorakh:
Teleports the player to random entities every 5-8 seconds


ChaosMod/Effects/db/Player/PlayerTpToEverything.cpp
```diff
@@ -0,0 +1,48 @@
...
+		allEntities.push_back(prop);
+	}
+
+	if (!allEntities.empty())
+	{
+		Entity randomEntity = allEntities[g_Random.GetRandomInt(0, allEntities.size() - 1)];
+		TeleportPlayer(GET_ENTITY_COORDS(randomEntity, true));
+	}
+
+	WAIT(g_Random.GetRandomInt(5000, 8000));
```
Last0xygen, line 38: Please change to a local variable and check on start of the onTick (see the airstrike effect)

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerTpToEverything.cpp
Goorakh: Fixed :)

PR #2714, opened 2022-02-16
Fixes a softlock with the mainthread-hook

Last0xygen:
Fixes a softlock if you activate an effect, that blocks the mainthread on death. This would result in the game not continuing / showing you restart/exit on missions.
-> only block the "mission_stat_watcher"-thread, to block any mission fails for distance / car destroyed
-> Disable the block if the screen is faded out

### DIFF
--- a/ChaosMod/Main.cpp
+++ b/ChaosMod/Main.cpp
@@ -1,6 +1,7 @@
 #include <stdafx.h>
 
 #include "Main.h"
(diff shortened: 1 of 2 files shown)


pongo1231:
The `mission_stat_watcher` block doesn't work since `strcmp` [returns 0](https://en.cppreference.com/w/cpp/string/byte/strcmp) if both strings match, so currently you are blocking every script thread that isn't `mission_stat_watcher` instead.

PR #2717, opened 2022-02-18
Changed the Effect-Config to JSON format

Last0xygen:
Keeping the old parsing method for LTS.
Update to the JSON format gets executed once the config-app gets opened.
This increases the filesize obviously but makes it way easier to maintain / extend.


pongo1231:
Implemented in comment above.

Issue #2661, opened 2022-01-10
[Bug] "Teleport To Random Location" destroys aircrafts


MoDerPo:
**Describe the bug**
The effect 'Teleport to random location' teleports the player to the ground level, even if they are in an aircraft which destroys it.

**Steps To Reproduce**
- press `ctrl`+`.` to pause the mod
- get in an aircraft
- press `ctrl`+`,` and search for "Teleport To Random Location"
- press `enter`
- kaboom

**Location (roughly)**
``ChaosMod/Effects/db/Player/PlayerTpController.cpp:204`` < there needs to be some check here.

pongo1231:
Already fixed; flying vehicles retain their current height when teleporting

PR #2727, opened 2022-02-23
Avoiding Teleporting Issues when in Aircraft

MoneyWasted:
Addresses pull request #2661 where if you are in a airplane or helicopter teleporting to a random location can cause the aircraft to get stuck and explode. This is a super crude fix and could be implemented into the TeleportPlayer function to handle it for all teleports.


DBWhale:
That would make it less chaotic

pongo1231:
Maybe it would be better to get the current Z offset from the ground and apply it to the new position instead?

Last0xygen (approved):

PR #2774, opened 2022-03-19
Added "Drought" Effect

MoneyWasted:
Does as you might think, it removes all the water from the map for a short period of time. View it in action [here](https://www.youtube.com/watch?v=-LhebMBqyr4)!


ChaosMod/Effects/db/Misc/MiscNoWater.cpp
```diff
@@ -0,0 +1,40 @@
...
+	Effect by MoneyWasted
+	Special Thanks to Parik for the Help :)
+*/
+
+#include <stdafx.h>
+
+uint64_t WaterPattern = Memory::FindPattern("? 6B C9 1C ? 03 0D ? ? ? ? 66 ? 03 C5 66 89 05 ? ? ? ?").Addr();
+uint64_t WaterPointer;
+
+std::vector<float> WaterHeights;
```
pongo1231, line 11: These should  be marked as static

ChaosMod/Effects/db/Misc/MiscNoWater.cpp
```diff
@@ -0,0 +1,40 @@
...
+		uint64_t Water = WaterPointer + (i * 0x1C);
+		*(float*)(Water + 0x14) = WaterHeights.at(i); // Restore Water
+	}
+	WaterHeights.clear();
+}
+
+static RegisterEffect registerEffect(EFFECT_MISC_WATER, OnStart, OnStop, nullptr, EffectInfo
+	{
+		.Name = "Drought",
+		.Id = "remove_water",
```
pongo1231, line 37: The id should follow the naming convention of the other effects (the `misc_` prefix is missing here)

pongo1231 (changes requested):

pongo1231:
There should be some fallback in case the pattern can't be found (e.g. by doing nothing in case `handle.IsValid()` is false).

Also the config app entry for the effect is missing.

ChaosMod/Effects/db/Misc/MiscNoWater.cpp
```diff
@@ -0,0 +1,59 @@
...
+	char unk2[4]; // 0x10
+	float Z;      // 0x14
+	uint Flags;   // 0x18
+};
+static_assert(sizeof(CWaterQuad) == 0x1C);
+
+static CWaterQuad* WaterQuads;
+static std::vector<float> WaterHeights;
+
+CWaterQuad* GetWaterQuads()
```
pongo1231, line 25: Again static is missing here :p

ChaosMod/Effects/db/Misc/MiscNoWater.cpp
```diff
@@ -0,0 +1,59 @@
...
+	{
+		WaterQuads[i].Z = WaterHeights.at(i); // Restore Water
+	}
+	WaterHeights.clear(); // Clear Storage Vector
+}
+
+static RegisterEffect registerEffect(EFFECT_MISC_WATER, OnStart, OnStop, nullptr, EffectInfo
+	{
+		.Name = "Drought",
+		.Id = "remove_water",
```
pongo1231, line 56: This also needs to be changed to `misc_remove_water`

pongo1231 (changes requested):

ChaosMod/Effects/db/Misc/MiscNoWater.cpp
```diff
@@ -0,0 +1,59 @@
...
+	uint Flags;   // 0x18
+};
+static_assert(sizeof(CWaterQuad) == 0x1C);
+
+static CWaterQuad* WaterQuads;
+static std::vector<float> WaterHeights;
+
+static CWaterQuad* GetWaterQuads()
+{
+	Handle handle = Memory::FindPattern("? 6B C9 1C ? 03 0D ? ? ? ? 66 ? 03 C5 66 89 05 ? ? ? ?");
```
pongo1231, line 27: Should be static so it only searches for that pattern the first time

ChaosMod/Effects/db/Misc/MiscNoWater.cpp
```diff
@@ -0,0 +1,59 @@
...
+	if (handle.IsValid())
+	{
+		return *handle.At(6).Into().Get<CWaterQuad*>();
+	}
+}
```
pongo1231, line 32: Should return nullptr if the pattern wasn't found

ChaosMod/Effects/db/Misc/MiscNoWater.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+static void OnStart()
+{
+	WaterQuads = GetWaterQuads();
```
pongo1231, line 36: Would be great to have a null check here (and for OnStop too) in case the pattern wasn't found so rather than the mod crashing the effect does nothing

pongo1231 (changes requested):

PR #2779, opened 2022-03-24
Add effect "Controller Experience"

Goorakh:
Offsets look inputs to simulate a "warped control stick".


ChaosMod/Effects/db/Player/PlayerWarpedControlStick.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+	SET_GAMEPLAY_CAM_RELATIVE_PITCH(GET_GAMEPLAY_CAM_RELATIVE_PITCH() - 1, 1.f);
+
+	for (int i = 0; i < 10; i++)
+	{
+		SET_GAMEPLAY_CAM_RELATIVE_HEADING(GET_GAMEPLAY_CAM_RELATIVE_HEADING() + (horizontalOffset / 10));
+		WAIT(0);
+	}
+
+	// TODO: Possibly offset the cursor position aswell? Windows API calls don't seem to work for that unfortunately, would probably have to find a memory address to override the position instead
```
MoneyWasted: Have you looked at [SetCursorLocation](https://docs.fivem.net/natives/?_0xFC695459D4D0E219) it directly calls the Windows API function for [SetCursorPos](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setcursorpos). Also keep in mind that this native can alter the cursor position even when focus is not on the game window. So be careful :)

Goorakh: I've already implemented this (the comment is wrong btw, it didn't work since it's parameters were pixels, while I had a float value <1, so the conversion would always end up passing an offset of 0, which of course did nothing), you're looking at an older commit, and yeah, it'll offset outside the game window (and possibly also when the game is minimized), shouldn't be too difficult to check for that though, as long as there's API calls for the window bounds and if the window is minimized it should be pretty simple.

MoneyWasted (commented):
Couldn't most of this code be reduced just by hooking WndProc? As that handles all the input along.

Goorakh:
> Couldn't most of this code be reduced just by hooking WndProc? As that handles all the input along.

It probably could, but I have no idea how to do that yet (nor that it was even possible). I might look into it, but otherwise I feel like the effect is fine as it is.

PR #2780, opened 2022-03-25
Added "Fleetwood" Effect

OnlyRealNubs:
Basically this effect replaces your vehicle with a dark green emperor. Similar to Lamar's. Sorry if this is a bad effect. Running out of ideas.


ChaosMod/Effects/db/Misc/MiscSelfieTime.cpp
```diff
@@ -0,0 +1,21 @@
+/*
+	Effect by OnlyRealNubs
+*/
+
+#include <stdafx.h>
+
+static void OnStart()
+{
+	DESTROY_MOBILE_PHONE();
```
MoneyWasted, line 9: This native causes soft locks during important missions that rely on phone interaction such as The Merryweather Heist (Offshore Variant). Please avoid using this native and find a way to wait for the phone call to be over.

MoneyWasted (changes requested):

ChaosMod/Effects/db/Misc/MiscSelfieTime.cpp
OnlyRealNubs: That wasn't supposed to be there. I might of typed it on accident. I will immediately remove it.

OnlyRealNubs: That whole file was accidentally included.

PR #2781, opened 2022-03-25
Added "Go Straight To Jail" effect.

OnlyRealNubs:
This effect spawns a police car with a cop in the driver seat. Then the script puts the player in the back of the police car. The cops is given an AI task to drive to the back of the police station.  If the player bails out of the car or the cop reaches its destination, the siren will turn off and both the cop and the police car are set as no longer needed.


OnlyRealNubs:
This effect name is reference to the Monopoly game's space "Go Straight To Jail, Do Not Pass Go"

ChaosMod/Effects/db/Misc/MiscGoToJail.cpp
```diff
@@ -0,0 +1,69 @@
...
+		SET_ENTITY_AS_NO_LONGER_NEEDED(&veh);
+		SET_ENTITY_AS_NO_LONGER_NEEDED(&policePed);
+	}
+	else if (GET_VEHICLE_PED_IS_IN(PLAYER_PED_ID(), true) != veh)
+	{
+		SET_VEHICLE_SIREN(veh, 0);
+		SET_ENTITY_AS_NO_LONGER_NEEDED(&veh);
+		SET_ENTITY_AS_NO_LONGER_NEEDED(&policePed);
+	}
+}
```
Reguas, line 59: OnTick wouldn't work with non-timed effects. What you're trying to do is to make a "semi-permanent" effect, which is currently not possible.

OnlyRealNubs: Ok. So do I put it in a while loop?

OnlyRealNubs: Ok. So we remove it.

PR #2789, opened 2022-03-28
Added "Teleport to Cayo Perico" to teleports list

OnlyRealNubs:
Honestly, this effect is a bit slow. Because it loads in the MP DLC maps. This means that stuff like the GTA Online casino and other MP DLC IPLs will be loaded in as well. When the player is teleported there, a plane is spawned so they can leave the island.


MoneyWasted:
Are you sure you need to load every MP IPL? I've gotten the island to load only with the island IPLs. Also from my research loading MP IPLs in single player is a hit or miss. Some load, some don't or sometimes it just crashes. Have you had any issues other than the long wait?

OnlyRealNubs:
The only way for me to get the island IPLs to load when I call to load them, is to enable the mp maps. I tried without and you would just fall in the water.

MoneyWasted:
> The only way for me to get the island IPLs to load when I call to load them, is to enable the mp maps. I tried without and you would just fall in the water.

[here](https://gist.github.com/MoneyWasted/f20f1c208ace8ac0e12aee7e6a08a32c) is the code for mine. Its directly from decompiled scripts. It crashes for me, but it might be useful for you to look at.

OnlyRealNubs:
Alright. When I get a chance tomorrow, I will look at the link you supplied and will make changes accordingly.

OnlyRealNubs:
I went over your code. And it I found what was making it crash. But further more you can't load MP IPLs without calling `ON_ENTER_MP`. Which will put the player into a blurry screen with it saying `Loading Online` in the bottom corner.

OnlyRealNubs:
Closes because is shit.

PR #2790, opened 2022-03-28
Added "Eject!!" Effect

OnlyRealNubs:
This effect ejects all peds from the current vehicle. Like from the mission Deep Inside, but not as glamorous and cool looking.


Last0xygen:
Imho too similar to „ragdoll all peds“, „everyone exits their vehicle“ and/or „launch all peds up“

OnlyRealNubs:
'Aight. It was made quickly anyways. I'm working on another effect right now.

PR #2796, opened 2022-03-29
Added effect "That's Stunning"

OnlyRealNubs:
This effect draw a stun gun bullet at the player. Making them fall to the ground. This has an unintentional side effect when you are in a vehicle that I decided to keep, when the player gets stunned inside a vehicle, they will be locked out of the vehicle controls. They cannot steer, brake, accelerate or exit. If the we're accelerating, they will be stuck doing that until they regain control.

It's. Like. Literally Stunning.


OnlyRealNubs:
Turns out the "duplicate file" wasn't the duplicate. It was the right file. The actual duplicate file is in the branch. I am fixing it right now.

OnlyRealNubs:
e6ad623

Added correct file to branch. Now when the player is stunned they will get a screen effect. The same one they get from the random event with Trevor.

ChaosMod/Effects/db/Player/PlayerStun.cpp
```diff
@@ -0,0 +1,30 @@
...
+*/
+
+#include <stdafx.h>
+
+std::string effectName = "Dont_tazeme_bro";
+
+static void OnStart()
+{
+	Vector3 pos1 = GET_OFFSET_FROM_ENTITY_IN_WORLD_COORDS(PLAYER_PED_ID(), 0, 1, 0);
+	Vector3 pos2 = GET_ENTITY_COORDS(PLAYER_PED_ID(), 1);
```
Last0xygen, line 12: get "PLAYER_PED_ID" once before that

ChaosMod/Effects/db/Player/PlayerStun.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+	Ped owner = GET_RANDOM_PED_AT_COORD(pos1.x, pos1.y, pos1.z, 50, 50, 50, -1);
```
Last0xygen, line 14: Why is this needed? You can just set "0" as the "owner" in "SHOOT_SINGLE_BULLET_BETWEEN_COORDS" (see the  Airstrike effect)

ChaosMod/Effects/db/Player/PlayerStun.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+	SHOOT_SINGLE_BULLET_BETWEEN_COORDS(pos1.x, pos1.y, pos1.z, pos2.x, pos2.y, pos2.z, 0, 0, 0x3656C8C1, owner, 1, 0, 100);
+	
+	ANIMPOSTFX_PLAY(effectName.c_str(), 0, 0);
+	WAIT(4600);
+	ANIMPOSTFX_STOP(effectName.c_str());
+}
+
+// Any of these functions can be omitted and either replaced with a `nullptr` or completely left out (default parameter values) in the `RegisterEffect` declaration
```
Last0xygen, line 23: comment can be removed

ChaosMod/Effects/db/Player/PlayerStun.cpp
```diff
@@ -0,0 +1,30 @@
...
+static RegisterEffect registerEffect(EFFECT_PLAYER_STUN, OnStart, nullptr, nullptr, EffectInfo
+	{
+		// These are always required, you may have to add more designators depending on your effect
```
Last0xygen, line 26: comment can be removed

ConfigApp/Effects.cs
```diff
@@ -345,11 +345,12 @@ public enum EffectType
             EFFECT_PLAYER_LAGGY_CAMERA,
             EFFECT_MISC_CLONE_ON_DEATH,
             EFFECT_SIDEWAYS_GRAVITY,
-            EFFECT_MISC_JUMPY_PROPS,
+            EFFECT_MISC_JUMPY_PROPS,
```
Last0xygen, line 348: unneeded spaces after the comma

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerStun.cpp
OnlyRealNubs: Had no idea I could do that. Thanks for pointing that out.

OnlyRealNubs:
@Last0xygen all fixed. Thanks for pointing out that I can set ped to 0 in `SHOOT_SINGLE_BULLET_BETWEEN_COORDS`

Last0xygen:
> @Last0xygen all fixed. Thanks for pointing out that I can set ped to 0 in `SHOOT_SINGLE_BULLET_BETWEEN_COORDS`

Nah, you missed my "PLAYER_PED_ID" comment 😅 
You can get that once and provide the variable to the two functions instead of getting it twice.
Otherwise it has been nicely adjusted 👍

Last0xygen (approved):

PR #2811, opened 2022-04-01
Added "Water Guns" effect

OnlyRealNubs:
After a lot of suffering, it finally works.


ChaosMod/PedsWaterGuns.cpp
```diff
@@ -0,0 +1,48 @@
...
+	for (Ped ped : GetAllPeds())
+	{
+		if (IS_PED_SHOOTING(ped))
+		{
+			USE_PARTICLE_FX_ASSET(particleDict);
+			int wep = GET_CURRENT_PED_WEAPON_ENTITY_INDEX(ped);
+			SET_PARTICLE_FX_SHOOTOUT_BOAT(1);
+			int ptfx = START_PARTICLE_FX_LOOPED_ON_ENTITY(particleName, wep, 0.f, 0, -0.15f, 0, 0, -90.f, 1, 0, 0, 0);
+			WAIT(300);
+			STOP_PARTICLE_FX_LOOPED(ptfx, 0);
```
Last0xygen, line 23: Please change the layout to something like with the flamethrower layout. no "wait(300)" as this would delay every following effect further and further.

Last0xygen (changes requested):

Last0xygen (changes requested):
water-animation rotation is not right

PR #2816, opened 2022-04-06
[Config App] Added ability for the player to add custom incompatible effects.

OnlyRealNubs:
I made some changes to the config app. These changes lets the player enter the config for the effect. They will see a Button that says "Incompatible with these effects". When they click the button, a new window will popup, where they can add and remove effects from the list. I had to make this a new window because there was no way to fit it inside the original window. When save is hit, the incompatible effects are stored by id in the `effects.ini` file. It was originally stored by their enum name. But was changed to id as it was more reliable. 

I added some code in `EffectConfig.h` to read the values from the `effects.ini` and add them to the `IncompatibleIds` vector.

image
image

I haven't done c# in a while. So please if you something that I can do to improve in the code. Please comment and let me know.


OnlyRealNubs:
This could be better in JSON format. Instead of storing them all in one string and splitting them by commas `,`. We could store them in a array.

Last0xygen:
Yeah, obviously, but this is 100% a conflict ;P

ConfigApp/EffectConfig_incompatibility.xaml.cs
```diff
@@ -0,0 +1,172 @@
...
+                if (selectedType == EffectType._EFFECT_ENUM_MAX) { return; }
+                ListData selectedData = effects[selectedType];
+                selectedData.inList = false;
+                effects[selectedType] = selectedData;
+                effectconf_custom_incompatibility_remove_list.Items.Remove(selectedName);
+                effectconf_custom_incompatibility_add_list.Items.Add(selectedName);
+                SortLists();
+            };
+
+            effectconf_custom_incompatibility_add_button.Click += (object sender, RoutedEventArgs e) =>
```
Last0xygen, line 120: Extracting those lambdas into separate functions help the readability of this code

OnlyRealNubs: The reason I made them lambdas is because I needed to access `effectData` and `effectInfo` variables. Not the best at c#. Is there a way to access them in a separate function?

OnlyRealNubs:
This has a few conflicts that I can't fix for some reason :/

PR #2825, opened 2022-04-11
Added "Play Random Cutscene" effect

reesercollins:
Plays a randomly chosen cutscene. I have preventing it from playing cutscenes while one is currently running as that softlocks the game.


OnlyRealNubs:
The only problem I see with this is that if the ped the player is currently using is in the cutscene (for example `Michael`) the ped will be teleported to the place of the cutscene.

reesercollins:
It doesn't seem to do that. The player is where you were before the effect when you skip the cutscene.

OnlyRealNubs:
Alright. After building and testing the effect You seem to be correct. I attempted the same thing a week ago. Except for me it wouldn't teleport the player back. Everything here is good. Make sure to put your name at the top the effect. I saw you left it blank.

OnlyRealNubs (approved):

rollschuh2282:
there are a couple cutscenes like the one for Michael knocking franklin out in repossessions if the player doesn´t drive to simeon and some other ones which are at a custom location depending on where the player is.
in the game files they are placed at 0.0.0 so under the map and R* plays them via script at the custom location.
maybe it would be useful to exclude those or script these specific ones at a set location to prevent the player wondering why they are under the map.
and maybe some more GTAO ones could maybe be added

reesercollins:
> there are a couple cutscenes like the one for Michael knocking franklin out in repossessions if the player doesn´t drive to simeon and some other ones which are at a custom location depending on where the player is. in the game files they are placed at 0.0.0 so under the map and R* plays them via script at the custom location. maybe it would be useful to exclude those or script these specific ones at a set location to prevent the player wondering why they are under the map. and maybe some more GTAO ones could maybe be added

There is a native function to play a cutscene at a specific location. Now I know why. I will make some changes later to fix that issue. As for the GTAO cutscenes, I haven't had a look, but do you happen to have a list of all online cutscene IDs?

OnlyRealNubs:
Also some cutscenes require a script to terminate them. So if the player doesn't know how to skip cutscenes, or for some reason they can't skip, they will be stuck like that forever.

OnlyRealNubs (changes requested):

ChaosMod/Effects/db/Player/PlayerPlayRandomCutscene.cpp
```diff
@@ -0,0 +1,468 @@
...
+			START_CUTSCENE(0);
+		} else {
+			Ped playerPed = PLAYER_PED_ID();
+			Vector3 playerPos = GET_ENTITY_COORDS(playerPed, false);
+			START_CUTSCENE_AT_COORDS(playerPos.x, playerPos.y, playerPos.z, 0);
+		}
+	}
+}
+
+// Any of these functions can be omitted and either replaced with a `nullptr` or completely left out (default parameter values) in the `RegisterEffect` declaration
```
OnlyRealNubs, line 461: You can remove the comments here.

reesercollins:
Feature already implemented

PR #2828, opened 2022-04-11
Added "Giants" effect

OnlyRealNubs:
Multiplies the scale of the all the peds by 3. Creates a custom camera so you can actually see your player. The legs do clip into the ground but I couldn't find a solution to it. Also sets the pitch low.

_"Stole"_ @Goorakh's scaling [code](https://github.com/Goorakh/ChaosModV/blob/e7ef637ed59167d2064852ed0d0b4b3c07b77824/ChaosMod/Memory/Vehicle.h#L185) from "Tiny Vehicles" but for peds.


Reguas:
There's ViewMatrix at https://github.com/gta-chaos-mod/ChaosModV/pull/2768 that helps reduce code duplication.
I also played around with this concept a while ago, and peds don't scale in vehicles unlike with Minions, so I thought it wasn't a great effect idea.

OnlyRealNubs:
> peds don't scale in vehicles

I think it's the animations that scale them down. So would need to find a way to scale them in the animation

OnlyRealNubs:
ok, so, I had them scaled in vehicles, which didn't turn out so great, it was very glitchy, but it all just adds to the chaos.

PR #2840, opened 2022-04-14
[W.I.P] Added effect "Gravity Guns"

OnlyRealNubs:
This effect gets the entity the player is aiming at. Then it attaches it to the players weapon. If the player stops aiming the entity will be detached. I set this as W.I.P because it is very minimal but still does something.


Reguas:
https://github.com/gta-chaos-mod/ChaosModV/pull/2284 🤔

OnlyRealNubs:
> #2284 🤔

That effect is more like grapple guns. It pulls the object towards the player. This effect let's the player hold the entity at the tip of their gun. Defying gravity.

Goorakh:
> > #2284 🤔
> 
> That effect is more like grapple guns. It pulls the object towards the player. This effect let's the player hold the entity at the tip of their gun. Defying gravity.

It's still a name conflict. 'Grapple Guns' would not be a better name for my effect, it would probably make players expect it to be some kind of grappling hook ability, which is not what the effect does. I ended up calling it 'Gravity Guns' as a vague reference to HL2 (Sticking to the same "theme" as 'Portal Guns').

OnlyRealNubs:
> > > #2284 🤔
> > 
> > 
> > That effect is more like grapple guns. It pulls the object towards the player. This effect let's the player hold the entity at the tip of their gun. Defying gravity.
> 
> It's still a name conflict. 'Grapple Guns' would not be a better name for my effect, it would probably make players expect it to be some kind of grappling hook ability, which is not what the effect does. I ended up calling it 'Gravity Guns' as a vague reference to HL2 (Sticking to the same "theme" as 'Portal Guns').

Ok. I fixed the name conflict. I though @Reguas referenced it as a duplicate. Sorry for the misunderstanding on my behalf.

OnlyRealNubs:
https://github.com/gta-chaos-mod/ChaosModV/pull/2840/commits/02aeae397f55a7056225a0684b0ef3e741310f32

You can now grab AI vehicles. You can now hit the shoot button to send an object flying through the air. Changed to RayCasting from the gameplay cam to get entities as they wouldn't get picked up properly sometimes with the old method. Switched to detecting input instead of getting the players aiming state. You no longer have to be constantly aiming to pickup an object. You press it, it scans, if there is an object it will pick it up for as long as you have he aim button down. Thanks to pongo for helping me with some errors. And MoneyWasted for the RayCasting.

OnlyRealNubs:
https://github.com/gta-chaos-mod/ChaosModV/pull/2840/commits/431d0ef03c8be290bdb84b4b60865b27c83537b4

Fixed object still being attached to you when you meant to let go. Changed grab distance and force applied to object. Changed the offset of the object. `CleanUp` function now gives the object gravity. The player now gets infinite ammo while the effect is active.

OnlyRealNubs:
Here a [link](https://vimeo.com/699890838) to a video showcasing the effect.

Last0xygen (changes requested):
Tested the effect, seems a bit "complicated", as you cant really see where you're aiming at before you're really aiming.
I would trigger the picking up also with the shooting button and block firing the weapon if you're picking up or firing an object :)

PR #2875, opened 2022-04-20
Added "Give Everyone A Flare Gun"

OnlyRealNubs:
If you shoot someone with a flare gun, they go up in flames. So I added flare guns. Might not be the best idea....


ChaosMod/Effects/db/Peds/PedsEveryoneWepController.cpp
```diff
@@ -127,4 +127,23 @@ static RegisterEffect registerEffect7(EFFECT_EVERYONE_BATTLEAXE, OnStartBattleAx
...
+);
+
+static void OnStartFlareGun()
+{
+	static const Hash flareGunHash = GET_HASH_KEY("WEAPON_FLAREGUN");
+
+	for (Ped ped : GetAllPeds())
+	{
+		GIVE_WEAPON_TO_PED(ped, flareGunHash, 9999, true, true);
+		SET_CURRENT_PED_WEAPON(ped, flareGunHash, true);
```
Last0xygen, line 139: the second "true" is "equip now", so the second native should not be necessary

OnlyRealNubs: I just copied the code from all the other give everyone effects. But I can change it, when I get back to my pc.

OnlyRealNubs:
Don't feel like resolving these merge conflicts on this effect

PR #2878, opened 2022-04-21
Added "FPS Counter" effect

OnlyRealNubs:
Displays the games FPS in a scaleform on the screen. The same scaleform that "Simeon Says" and "Need For Speed" uses.

For me displays from 20-30 fps average. I have a crappy GPU.


ConfigApp/Effects.cs
```diff
@@ -664,6 +665,7 @@ public enum EffectType
             {EffectType.EFFECT_MISC_WATER, new EffectInfo("Drought", EffectCategory.MISC, "misc_remove_water", true)},
             {EffectType.EFFECT_VEHS_PROP_MODELS, new EffectInfo("Prop Cars", EffectCategory.VEHICLE, "vehs_prop_models", true)},
             {EffectType.EFFECT_VEHS_TINY, new EffectInfo("Tiny Vehicles", EffectCategory.VEHICLE, "vehs_tiny", true)},
+            {EffectType.EFFECT_PLAYER_FPS,, new EffectInfo("FPS Counter", EffectCategory.PLAYER, "player_fps", true, false)}
```
Last0xygen, line 668: double comma, fails to compile

ChaosMod/Effects/db/Player/PlayerFPS.cpp
```diff
@@ -0,0 +1,34 @@
...
+	scaleform = REQUEST_SCALEFORM_MOVIE("MP_BIG_MESSAGE_FREEMODE");
+	while (!HAS_SCALEFORM_MOVIE_LOADED(scaleform))
+	{
+		WAIT(0);
+	}
+}
+
+static void OnTick()
+{
+	float fps = (1000 / GET_FRAME_TIME()) / 1000;
```
Last0xygen, line 19: "0.x Gamespeed" multiplies this value. Also showing the weapon wheel or holding alt increases the counter times 10 or something like that

Last0xygen (changes requested):

OnlyRealNubs:
This doesn’t make a good effect imo. So i’m just going to close it.

PR #2890, opened 2022-04-25
Added "Papers Please" effect

OnlyRealNubs:
Displays the papers please booth with a returning character from a previous GTA game. You have only the option to detain them. (I had to draw a custom cursor because the booth texture overlaid it.) This uses external textures (png). There are three pngs, each ~120 kb. They are stored in a folder named after the effect id in a data folder in the chaosmod folder. Similar to the way "Google Translated" stored the label dict.  If the player doesn't click the detain button in time, it will auto end.


OnlyRealNubs:
Also please don't mention the "*good*" job I did photoshopping.

PR #2894, opened 2022-04-28
Added "Effect Of Choice" effect

OnlyRealNubs:
Opens the debug menu and let's the player select an effect that they want. Or they could just close it.


OnlyRealNubs:
trash, sloppy, poorly made

PR #2900, opened 2022-04-30
Added "6 Wanted Stars" effect

OnlyRealNubs:
Gives the player 6 wanted starts (actually five). The way this works is with fake wanted levels. So this is a semi-permanent effect, because fake wanted levels don't go away on their own, even when you die or get busted.


OnlyRealNubs:
image

MoneyWasted:
I find it funny that there is even the ability to set the wanted level to 6. New 6 star DLC soon?

OnlyRealNubs:
> I find it funny that there is even the ability to set the wanted level to 6. New 6 star DLC soon?

It's for r* to override wanted levels in stuff like missions. Then they dispatch their own cps manually. Never seen them use 6 stars though. 🤔

MoneyWasted:
> > I find it funny that there is even the ability to set the wanted level to 6. New 6 star DLC soon?
> 
> 
> 
> It's for r* to override wanted levels in stuff like missions. Then they dispatch their own cops manually. Never seen them use 6 stars though. 🤔 

Like I find it weird that the native would even register 6 stars like I know in the other GTA games it would just set the wanted level to what ever the highest was and ignore the value.

OnlyRealNubs:
> Like I find it weird that the native would even register 6 stars like I know in the other GTA games it would just set the wanted  level to what ever the highest was and ignore the value.

R* re-using code? 😏

OnlyRealNubs:
[53072a2](https://github.com/gta-chaos-mod/ChaosModV/pull/2900/commits/53072a2f39e6821ad7fcb8c6f2aae4f115b6a3e6)

Added a wanted level multiplier, so it actually feels like 6 stars would. Also increased the duration to 180 seconds.

MoneyWasted:
Maybe you could also make it spawn the military for a true 6-star experience!

ChaosMod/Effects/db/Player/PlayerWantedController.cpp
```diff
@@ -1,5 +1,30 @@
...
+{
+	Player player = PLAYER_ID();
+
+	SET_FAKE_WANTED_LEVEL(6);	
+	
+	SET_PLAYER_WANTED_LEVEL(player, 5, false);
+	SET_PLAYER_WANTED_LEVEL_NOW(player, false);
+
+	g_pEffectDispatcher->OverrideEffectName(EFFECT_6_STARS, EFFECT_6_STARS);
+	WAIT(60000);
```
MoneyWasted: Also, if it is going to be a timed effect use the OnStop and OnTick for handling the time. It also allows for other improvements as mentioned above :)

MoneyWasted (changes requested):

ChaosMod/Effects/db/Player/PlayerWantedController.cpp
OnlyRealNubs: Yup. I will surely make it a timed effect. OnStop should I remove the wanted level, or just give them a 5 Start instead?

PR #2907, opened 2022-05-02
Added Version Check

OnlyRealNubs:
Checks the current version to the up-to-date version at https://gopong.dev/chaos/version.txt. If it is outdated, a message will appear notifying the user that the mode is outdated. Uses [httplib ](https://github.com/yhirose/cpp-httplib) library for http requests.

Note: At the top of `stdafx.h` is defined the current mod version (`MOD_VERSION`). That pre-processor has to change when the mod version changes.


OnlyRealNubs:
It was hard to find a functioning http library that didn't require 15 million other libraries.

MoneyWasted:
I have an even smaller http lib if you want to check it out. :P

OnlyRealNubs:
> I have an even smaller http lib if you want to check it out. :P

Absolutely. Please DM it to me.

PR #2921, opened 2022-05-07
Adding Super Simple HTTP Implementation

MoneyWasted:
Can be used to make GET, POST, PUT and DELETE requests. Conveniently fits into 2 small files that combined are under 8 kilobytes and it's all done without the need for bloated libraries. This code is from a very old project of mine. I tried my best to clean and optimize the code but if there is anything that can be improved, please let me know and I'll make the needed changes.

Here is a simple example of how to use it. 
```cpp
// The root domain
const std::wstring domain = L"gopong.dev";

// We need a content header just use JSON because it is universal
const std::wstring requestHeader = L"Content-Type: application/json";

// HTTPS is 443, if you are requesting a non-HTTPS site use port 80
int port = 443;

// Are we using HTTPS? If not, this should be false.
bool https = true;

// Create our request object
HTTP req(domain, port, https);

// Don't forget the response object!
HttpResponse response;

// Create a GET Request with our new object
req.Get(L"/chaos/version.txt", requestHeader, response);

// For testing we just log to a console but 'response.text' is the std::string for the content from the site.
std::cout << response.text << std::endl;

// Reset the response if we are going to use this again to avoid overlapping content.
response.Reset();
```


OnlyRealNubs:
Could be used with #2907.

PR #2929, opened 2022-05-09
Added "Player Has No Collision" effect

OnlyRealNubs:
What the title says. Player can't collide with cars, peds objects etc. The player won't fall through the map. And if the player is in a vehicle that vehicle will have no collision as well. Uses 3 for loops, since there is no function to get all entities.


Last0xygen:
Duplicate ;)
https://github.com/gta-chaos-mod/ChaosModV/pull/333

OnlyRealNubs:
😢

PR #2933, opened 2022-05-11
Chaos Mod Installer

OnlyRealNubs:
Because some people are way to retarded to install it properly. I actually found it confusing to install for FiveM.

Simple installation instructions.

image
image

ChaosMod zip would look like this. With `bin` folder containing the chaosmod folder and `ChaosMod.asi`.
image

Let me know what you think 😃


OnlyRealNubs:
I do have future plans, where we would just ship the installer. And it would fetch the latest version files from the github.

PR #2935, opened 2022-05-11
Added "Re-Invoke Previous Effects" meta effect

OnlyRealNubs:
Dispatches the previous 5 effects that were dispatched. That simple.

If you have anything to add, please let me know. 😄


ChaosMod/Components/EffectDispatcher.cpp
```diff
@@ -566,6 +569,16 @@ void EffectDispatcher::ClearMostRecentEffect()
 	}
 }
 
+std::vector<RegisteredEffect*> EffectDispatcher::GetRecentEffects(int distance)
```
pongo1231, line 572: Nitpick but this could be marked as a const function as it doesn't modify any fields (`const` at the end of the line, also in the header)

ChaosMod/Components/EffectDispatcher.cpp
```diff
@@ -566,6 +569,16 @@ void EffectDispatcher::ClearMostRecentEffect()
...
+{
+	std::vector<RegisteredEffect*> temp;
+	for (int i = m_rgDispatchedEffectsLog.size() - 1; i --> distance;)
```
pongo1231, line 575: Please just do `i > distance; i--` instead, it's going to be more clearer to readers.

Also make sure to check if you haven't reached the end of the list (i < 0)

ChaosMod/Components/EffectDispatcher.cpp
```diff
@@ -566,6 +569,16 @@ void EffectDispatcher::ClearMostRecentEffect()
...
+	{
+		temp.emplace_back(m_rgDispatchedEffectsLog.at(i));
```
pongo1231, line 577: Maybe it should check if an effect is already in the list and ignore it (as you can't have 2 instances of an effect running simultaneously).

ChaosMod/Effects/Effect.h
```diff
@@ -44,6 +44,11 @@ struct RegisteredEffect
 		return m_EffectIdentifier == effectIdentifier;
 	}
 
+	EffectIdentifier GetIndentifier()
```
pongo1231, line 47: Return a const reference instead here as there's no reason to return a copy of EffectIdentifier, can also be marked as a const function.

ChaosMod/Effects/db/Meta/MetaReInvoke.cpp
```diff
@@ -0,0 +1,23 @@
+/*
+	Effect By OnlyRealNubs
+*/
+
+#include "stdafx.h"
+
+static void OnStart()
+{
+	EffectDispatcher* disp = GetComponent<EffectDispatcher>();
+	std::vector<RegisteredEffect*> eff = disp->GetRecentEffects(5);
```
pongo1231, line 10: Should accept a const reference here (copy elision should kick in here but still)

ChaosMod/Effects/db/Meta/MetaReInvoke.cpp
```diff
@@ -0,0 +1,23 @@
...
+	for (int i = 0; i < eff.size(); i++)
```
pongo1231, line 11: You can just use `for (RegisteredEffect* effect : eff)` instead

ConfigApp/Effects.cs
```diff
@@ -674,6 +675,7 @@ public enum EffectType
             {EffectType.EFFECT_MISC_WATER, new EffectInfo("Drought", EffectCategory.MISC, "misc_remove_water", true)},
             {EffectType.EFFECT_VEHS_PROP_MODELS, new EffectInfo("Prop Cars", EffectCategory.VEHICLE, "vehs_prop_models", true)},
             {EffectType.EFFECT_VEHS_TINY, new EffectInfo("Tiny Vehicles", EffectCategory.VEHICLE, "vehs_tiny", true)},
+            {EffectType.EFFECT_META_EFFECT_REINVOKE, new EffectInfo("Re-Invoke Previous Effects", EffectCategory.META, "meta_re_invoke", false)},
```
pongo1231, line 678: Again a nitpick but no reason to explicitly mark the timed flag as `false` here

pongo1231 (changes requested):

ChaosMod/Components/EffectDispatcher.cpp
```diff
@@ -566,6 +574,18 @@ void EffectDispatcher::ClearMostRecentEffect()
 	}
 }
 
+std::vector<RegisteredEffect*> EffectDispatcher::GetRecentEffects(int distance) const
+{
+	std::vector<RegisteredEffect*> temp;
+	for (int i = m_rgDispatchedEffectsLog.size() - 1; i > distance && i > 0; i--)
+	{
+		RegisteredEffect* regeff = m_rgDispatchedEffectsLog.at(i);
+		if (std::count(temp.begin(), temp.end(), regeff)) continue;
```
pongo1231, line 583: Use `std::find` instead (more efficient).

Additionally, if a duplicate effect was found it will still decrement `i`, meaning the amount of returned effects != `distance`. Instead of checking if `i` reached `distance` in the for loop declaration, it might be wiser to manually keep track of the amount (e.g. checking if it added an effect to `temp` and decrementing `distance` if so, until `distance` reaches <= 0)

pongo1231 (changes requested):

PR #2936, opened 2022-05-11
Added 3 new effects

OnlyRealNubs:
Added:

`Freight Friends` (Trains)
`Dozer Dudes` (Bull dozers)
~`Mower Mates` (Lawn Mowers)~

(Like the Bus Bois/Scooter Brothers effects)

This also puts all of those effects, into on file called `PedsVehicleController.cpp`


OnlyRealNubs:
Turns out mower mates is already an effect. So I credited the person who created it.

PR #2947, opened 2022-05-13
Added "Prop Land" effect

OnlyRealNubs:
Prop Hunt & Prop Vehicles in one. Credit @Goorakh for prop code, with some modifications made to it.


Last0xygen:
i don't think we need a "combination-effect" for both

OnlyRealNubs:
Alright. That’s fine. Thanks for the feedback.

PR #2996, opened 2022-05-25
Implemented debug shortcuts

OnlyRealNubs:
#### If the build configuration is set to debug:
You can set a debug shortcut for effects with `.DebugShortcutKeycode` in EffectInfo. This will override shortcuts read from the config file (if any), with the specified KeyCode. This saves time having to scroll through the debug menu.  I was originally going to do the same with the lua scripts (making the member only functional when using a debug build), but people writing lua scripts, don't necessarily have a debug build.

NOTE: The member doesn't exist in Release configuration.

### DIFF
--- a/ChaosMod/Effects/EffectsInfo.h
+++ b/ChaosMod/Effects/EffectsInfo.h
@@ -15,7 +15,7 @@ struct EffectInfo
 	bool IsTimed         = false;
 	bool IsShortDuration = false;
 #ifdef _DEBUG
-	int DebugShortcutKeycode;
+	int DebugShortcutKeycode = 0;
 #endif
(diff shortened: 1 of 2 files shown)


ChaosMod/Effects/EffectsInfo.h
```diff
@@ -14,6 +14,9 @@ struct EffectInfo
 	const char *Id;
 	bool IsTimed         = false;
 	bool IsShortDuration = false;
+#ifdef _DEBUG
+	int DebugShortcutKeycode;
```
pongo1231, line 18:
```suggestion
	int DebugShortcutKeycode = 0;
```

PR #3000, opened 2022-05-25
Added "Low Resolution" effect

OnlyRealNubs:
The screen in scaled down and centered.

image


Reguas:
It could be combined with DVD Screensaver

OnlyRealNubs:
> It could be combined with DVD Screensaver

Ok, I will make it incompatible.

OnlyRealNubs:
> You could also lower the bitrate (or whatever is name of the thing that make the screen pixelated) saying from personal experience as a shitty pc user trying to play GTA V 😂

Honestly think that just might make it unplayable, since the resolution is already lowered and then lowering the bit rate would just make it blurry.

ReiDaTecnologia:
> > You could also lower the bitrate (or whatever is name of the thing that make the screen pixelated) saying from personal experience as a shitty pc user trying to play GTA V 😂
> 
> Honestly think that just might make it unplayable, since the resolution is already lowered and then lowering the bit rate would just make it blurry.

Welcome to my life brother 😔

PR #3001, opened 2022-05-26
Added "Sticky Collision" effect

OnlyRealNubs:
When a vehicle collides, it will be frozen, as if it stuck to the surface it collided to. If collided with a vehicle, said vehicle will also be stuck.

(Actual name "Vehicles Have Sticky Collision". Shortened for title)

### DIFF
--- a/ChaosMod/Effects/db/Vehs/VehsStickyCollide.cpp
+++ b/ChaosMod/Effects/db/Vehs/VehsStickyCollide.cpp
@@ -1,0 +1,41 @@
+/*
+	Effect By OnlyRealNubs
+*/
+
+#include "stdafx.h"
+
+std::vector<Vehicle> vehs;
+
+static void OnStop()
+{
+	for (Vehicle& veh : vehs)
+	{
+		FREEZE_ENTITY_POSITION(veh, false);
+	}
+}
+
+static void OnTick()
+{
+	for (Vehicle veh : GetAllVehs())
+	{
+		if (true)
+		{
+			if (HAS_ENTITY_COLLIDED_WITH_ANYTHING(veh))
+			{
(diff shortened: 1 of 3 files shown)


TerrizX:
What about "Magentic Vehicles" or "Magnets"

ChaosMod/Effects/db/Vehs/VehsStickyCollide.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+static void OnTick()
+{
+	for (Vehicle veh : GetAllVehs())
+	{
+		if (true)
+		{
+			if (HAS_ENTITY_COLLIDED_WITH_ANYTHING(veh))
+			{
+				LOG("COLLIDED");
```
Last0xygen, line 25: Please Remove logs before submitting ;)

ChaosMod/Effects/db/Vehs/VehsStickyCollide.cpp
```diff
@@ -0,0 +1,41 @@
...
+		FREEZE_ENTITY_POSITION(veh, false);
+	}
+}
+
+static void OnTick()
+{
+	for (Vehicle veh : GetAllVehs())
+	{
+		if (true)
+		{
```
Last0xygen, line 22: unneeded if

Last0xygen (changes requested):

ChaosMod/Effects/db/Vehs/VehsStickyCollide.cpp
OnlyRealNubs: I don’t know why thats there 🫤

PR #3005, opened 2022-05-26
Some more hacking phrases

TheRealAyCe:


OnlyRealNubs:
Imo, I would add more than just more phrases. Like maybe make the effect more efficient, and then *sneak* them in there lol

Last0xygen:
Why create a new PR? @TheRealAyCe

TheRealAyCe:
> Imo, I would add more than just more phrases. Like maybe make the effect more efficient, and then _sneak_ them in there lol

Yeah, it's a bit lame. :P

> Why create a new PR? @TheRealAyCe

The other one had wrong spaces and modified the trailing line, which is unneccessary.

PR #3023, opened 2022-05-31
Added controller rumble to intense effects

OnlyRealNubs:
Added controller rumble (vibration) to intense effects like "Doomsday", "Earthquake" and "Black Hole".

Rumble doesn't need to be set each frame, but it is due to the game automatically stopping it when it rumbles the controller.

Uses [XInput](https://docs.microsoft.com/en-us/windows/win32/xinput/xinput-game-controller-apis-portal), because it is better than the `SET_PAD_SHAKE` native, as you can set different speeds for both left and right motors. Doesn't have the new rumble with the Xbox Series X|S controllers, only left and right.

### DIFF
--- a/ChaosMod/Util/XInput.h
+++ b/ChaosMod/Util/XInput.h
@@ -1,0 +1,46 @@
+#include "stdafx.h"
+
+#include <Xinput.h>
+
+namespace XInput
+{
+	//Doesn't include the new Xbox Series X|S rumbles
+	inline void SetControllerRumble(DWORD controllerId, int leftMotorSpeed, int rightMotorSpeed)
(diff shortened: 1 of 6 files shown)


ChaosMod/Util/XInput.h
```diff
@@ -0,0 +1,46 @@
+#include "stdafx.h"
```
pongo1231, line 1:
```suggestion
#pragma once

#include <stdafx.h>
```

pongo1231 (changes requested):
XInput lib also needs to be added to the CMake file, otherwise LGTM

PR #3030, opened 2022-06-01
Improved "Fake Death" effect

OnlyRealNubs:
## Current Improvements
- Now says "Mission Failed" when on a mission, followed by a random message and "[CHARACTER] died." 
- If the player bails out of the vehicle when it is being detonated, it will instead wait a while then skip to the cleanup section. Originally, it would say wasted despite the player not having any logical reason to be "dead".


ChaosMod/Effects/db/Player/PlayerFakeDeath.cpp
```diff
@@ -29,8 +28,29 @@ static int lastModeTime               = 0;
...
+	case 2608926626: // Trevor
+		return "Trevor";
+	default: // default
+		return "You";
+	}
+}
+
+static void OnStart()
+{	
+	AUDIO::REQUEST_SCRIPT_AUDIO_BANK("OFFMISSION_WASTED", 0, -1);
```
kolyaventuri, line 49: Is there an equivalent on-mission bank we could use so the sounds line up?

ChaosMod/Effects/db/Player/PlayerFakeDeath.cpp
```diff
@@ -196,16 +228,25 @@ static void OnStart()
...
+			if (GET_MISSION_FLAG())
+			{
+				SCALEFORM_MOVIE_METHOD_ADD_PARAM_PLAYER_NAME_STRING("~r~mission failed");
+				SCALEFORM_MOVIE_METHOD_ADD_PARAM_PLAYER_NAME_STRING(std::string(GetPlayerName()+" Died. " + ms_rgTextPairs[iChosenIndex]).c_str());
+			}
+			else
+			{
+				SCALEFORM_MOVIE_METHOD_ADD_PARAM_PLAYER_NAME_STRING("~r~wasted");
+				SCALEFORM_MOVIE_METHOD_ADD_PARAM_PLAYER_NAME_STRING(ms_rgTextPairs[iChosenIndex]);
+			}
```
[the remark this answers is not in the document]
OnlyRealNubs: I will see what I can do. Also thanks for the reviews, helps a lot. 🙃

ReiDaTecnologia:
I have a suggestion: maybe add "Ignite Player" as well? on foot it kinda works well (make the fire stop when it reaches 1 health point then re-heal player and show "Fake Death". Would be a good idea to check if the player aren't invincible as it could interfere with the effect.) 
In a vehicle it is a bit different:
"Ignite Player" normally just put the vehicle on fire and give you a good amount of time to stop the vehicle, leave it and take your distance (or you can just jump off the vehicle), you could lock the player inside of the vehicle, put the vehicle on fire for a short amount of time then fake blow it up.

SeanMirrsen:
> I have a suggestion: maybe add "Ignite Player" as well? on foot it kinda works well (make the fire stop when it reaches 1 health point then re-heal player and show "Fake Death". Would be a good idea to check if the player aren't invincible as it could interfere with the effect.) In a vehicle it is a bit different: "Ignite Player" normally just put the vehicle on fire and give you a good amount of time to stop the vehicle, leave it and take your distance (or you can just jump off the vehicle), you could lock the player inside of the vehicle, put the vehicle on fire for a short amount of time then fake blow it up.

I wouldn't futz around with it. Make it simpler.

On foot, for the duration of the fakery the player catches fire and flails around, but their health doesn't decrease (they are made immune to fire). When the fakery ends and "fake death" effect name is properly displayed, some seconds later, the fire is put out and the player stands up and continues as normal.

In a vehicle, the effect works exactly the same as the Detonate Vehicle fake effect, with the vehicle catching fire and, if the player doesn't leave it, exploding - and returning to normal afterward.

ChaosMod/Effects/db/Player/PlayerFakeDeath.cpp
```diff
@@ -141,8 +159,23 @@ static void OnStart()
...
+							if (!IS_PED_IN_VEHICLE(playerPed, veh, false))
+							{
+								if (GET_DISTANCE_BETWEEN_COORDS(vehCoords.x, vehCoords.y, vehCoords.z, plrCoords.x, plrCoords.y, plrCoords.z, false) > 1.5)
+								{
+									GetComponent<EffectDispatcher>()->OverrideEffectNameId("player_fakedeath",
+									                                                       fakeEffectId);
+									WAIT(2000);
+
+									currentMode = FakeDeathState::cleanup;
+									goto SKIP_TO_CLEANUP; //eww, a goto statement
```
Last0xygen, line 176: eww, a goto statement

Last0xygen (changes requested):

ChaosMod/Effects/db/Player/PlayerFakeDeath.cpp
OnlyRealNubs: What you want me to do about it, goto is the new meta

Last0xygen: goto is ugly, errorprone and not something you should use :)

OnlyRealNubs: What you want me to use?

OnlyRealNubs: I’ll fix this when I can.

OnlyRealNubs: All fixed 👍

PR #3031, opened 2022-06-01
Implemented Unstable Effects

OnlyRealNubs:
### Unstable effects, what are they?
Unstable effects are any effect that have a high chance of crashing. And are deemed unstable by contributors.
### What are the unstable effects?
Currently, the unstable effects are...
```
Random Traffic
Add Max Upgrades To Every Vehicle
Add Random Upgrades To Every Vehicle
Vehicle Rain
Spawn Random Vehicle
Replace Current Vehicle
```
### What does this do?
This implements a way we can specify when registering an effect, whether or not it is stable or not, by default it is marked as stable. Unstable effects will not be parsed unless the user enables them in the config app. (Default is off). When the user is in the configuration window of an effect, they will be told whether or not it is stable. (i.e "Yes or "No") 

#### Images
image
image


OnlyRealNubs:
This could temporarily fix the crashing of effects like "Random Traffic" with the stub vehicles.

PR #3034, opened 2022-06-01
Improved "Lock Camera" effect

OnlyRealNubs:
## Current Improvements
- You can now aim (only left and right)
- Fix the jitteriness of the camera when trying to move it

### DIFF
--- a/ChaosMod/Effects/db/Player/PlayerLockCamera.cpp
+++ b/ChaosMod/Effects/db/Player/PlayerLockCamera.cpp
@@ -4,12 +4,6 @@
 
 #include <stdafx.h>
 
-static void OnStop()
-{
-	ENABLE_CONTROL_ACTION(0, 1, true);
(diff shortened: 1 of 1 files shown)


ChaosMod/Effects/db/Player/PlayerLockCamera.cpp
```diff
@@ -1,23 +1,32 @@
...
 */
 
 #include <stdafx.h>
 
+static void OnStop()
+{
+	ENABLE_CONTROL_ACTION(0, 1, true);
+	ENABLE_CONTROL_ACTION(0, 2, true);
+}
+
```
pongo1231, line 12: I don't think that's necessary here? (since DISABLE_CONTROL_ACTION only applies to the current frame iirc)

OnlyRealNubs: You’re correct

PR #3045, opened 2022-06-06
Added "Alien Invasion" effect

Rylxnd:
Similar to zombies with the spawns, but when the effect starts a UFO is spawned above the player. Then when an alien is spawned, the UFO goes to the spawn location, and "drops off" the alien.

https://user-images.githubusercontent.com/91900600/172184014-f536baee-824a-4a58-87fb-c403f5d34507.mp4

### DIFF
--- a/ChaosMod/Effects/db/Peds/PedsAliens.cpp
+++ b/ChaosMod/Effects/db/Peds/PedsAliens.cpp
@@ -4,7 +4,6 @@
 
 #include <stdafx.h>
 
-#include <chrono>
 #include "Util/Peds.h"
 
 static std::list<Ped> m_aliens;
@@ -17,7 +16,6 @@ static const std::vector<Hash> m_weapons =
 
 #pragma region UFO 
 static const Hash ufoModel = 0xB467C540;
-static const double beamTime  = 1200;
 static Object ufo;
 static bool ufoBusy     = false;
 #pragma endregion


ReiDaTecnologia:
jesus fucking christ man, this is insta-death and worse, it is a timed effect meaning the player would die repeatedly until the fail-safe kicked in...
I think would be better if only one alien spawned and the "alien" weapons would be choosen randomly for it (the up atomizer having higher chance of being choosen cuz it is aren't as op as the lazer smg and lazer minigun), after the alien died, a new alien would spawn (maybe with some seconds between each spawn?) and continue until the effect ended.
Does the ufo follow the player? and what about the alien spawning? does it use the alien spawn effect from Michael's side-mission "Grass-Roots"?

OnlyRealNubs:
> Does the ufo follow the player? and what about the alien spawning?

It does follow the player, because the aliens are spawned around the player. So if the player moved locations, the ufo and alien spawns move with it.

> does it use the alien spawn effect from Michael's side-mission "Grass-Roots"?

Originally, I had it play a ptfx. But ultimately I removed it due to the fact that it wasn't really noticeable, as the aliens usually spawned where you couldn't see them.

> Would be cool too if instead of the ufo just popping out of existence, he would just move extremely fast up wards and despawn out of the player's vision (Like it does in the GTAO's Alien Tattoo easter egg)

Would be a nice addition. 👍

ReiDaTecnologia:
> > does it use the alien spawn effect from Michael's side-mission "Grass-Roots"?
> 
> Originally, I had it play a ptfx. But ultimately I removed it due to the fact that it wasn't really noticeable, as the aliens usually spawned where you couldn't see them.

afaik, the effect was made for that reason, for the aliens to spawn in the player vision's without looking like it popped into existence but up to you man <3

OnlyRealNubs:
[4f489c3](https://github.com/gta-chaos-mod/ChaosModV/pull/3045/commits/4f489c3296f329221a322ed60bd9a2ae48e1d90b)

- Fixed PTFX Not playing when an alien spawns in.
- UFO now comes from and leaves towards the bottom of the map. Instead of just appearing and disappearing. 
- Decreased the number of aliens, decreased their damage and lowered their fire-rate. To prevent continuous death.
- Aliens are now nicely removed when the effect is finished.
- Fixed UFO getting despawned by the game.
- UFO now Slowly Rotates.

PR #3062, opened 2022-06-12
Added "Remove Waypoint" effect

OnlyRealNubs:
Removes the active waypoint.


RosyMiranto18:
What about the mission objectives though? Since i don't think they could be deleted. Can this be done with "Hide the Waypoint" in case of those?

ChaosMod/Effects/db/Misc/MiscNoWaypoint.cpp
```diff
@@ -0,0 +1,18 @@
+/*
+	Effect By OnlyRealNubs
+*/
+
+#include "stdafx.h"
```
pongo1231, line 5:
```suggestion
#include <stdafx.h>
```

pongo1231 (changes requested):

PR #3178, opened 2022-07-25
Added basic ability to limit Twitch voting

LukasHeinzl:
Config app updated with an additional text field that lets you enter Twitch usernames.

The voting proxy now only counts votes from those users unless the field is empty.


pongo1231:
Thanks!

PR #3197, opened 2022-08-02
quick update for 2699 to fix failsafe and mp vehicle bypass

holysnipz:
first time doing this stuff, totally have 0 clues what I'm doing.

only tested on Steam 2699

[MP vehicle in SP](https://streamable.com/on2oo3)
[Failsafe](https://streamable.com/uzz6lt)


pongo1231:
Thanks for the PR! I pushed the updated pattern, however we'll update the offset at runtime instead to ensure compatibility with older game builds.

PR #3200, opened 2022-08-03
Prevents possible softlock where player ends up in a train

SRG27:
Effects "Set Everyone into Random Vehicles", "Set Player into Closest Vehicle" and "Set Player into Random Vehicle" may put the player in a train, which would cause a softlock because the player cannot leave the train.

This PR prevents the above-mentioned softlock.


pongo1231:
I'm not sure if we want this behavior. While this does lead to a softlock it also has a comedic element to it and the mod usually frees the player from that situation rather quickly one way or another.

PR #3270, opened 2022-10-03
Replaced the existing shortcut picker box with one with more options

SRG27:
Replaced the existing box for picking shortcuts with a text box that allows the player to enter custom shortcuts.

**Possible Issues:** 
Labels for keys `[]\';/.,-= (on US keyboard layout) are displayed as "OemX".


pongo1231:
Rebased it on master and refactored some small bits, seems fine now. Will merge for now regardless of the "OemX" issue.

PR #3278, opened 2022-10-06
Fake Crash Update

NeverKnowDev:
Fake Crash will now minimize the GTA window after the sleep time ends.


ReiDaTecnologia:
I think it would be funnier if the taskbar's icon of GTA V also disappeared but i think it's going too far for a troll effect

NeverKnowDev:
@ReiDaTecnologia Hiding the taskbar icon, showing an error messagebox is all possible but yeah it would be a bit too far for an effect. Tbh I will close this request as minimizing the window is already going too far and outside the game

PR #3337, opened 2023-01-08
Improve Blackout

rollschuh2282:
Decided to add a Extra Timecycle modifer just Like R* does it


pongo1231:
This would collide with the other timecycle mod effects and would thus need to be assigned to the Timecycle effect category

PR #3366, opened 2023-02-18
Added effect "Not Menendez!"

Reguas:
When anybody dies, someone else shouts "He's got Menendez!"

### DIFF
--- a/ChaosMod/Memory/Audio.h
+++ b/ChaosMod/Memory/Audio.h
@@ -1,0 +1,55 @@
+#pragma once
+
+#include "../Util/Logging.h"
+#include "Handle.h"
+#include "Memory.h"
+#include "Util/Natives.h"
+
+#include <scripthookv/inc/main.h>
+
+namespace Memory
+{
+	typedef uintptr_t(__cdecl PlayAmbientSpeech)(uintptr_t a1, unsigned int speechNameHash, const char *speechParam,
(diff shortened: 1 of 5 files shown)


pongo1231:
Any reason the native PLAY_AMBIENT_SPEECH_FROM_POSITION_NATIVE isn't used? Also it's a better fit for the Peds effect category rather than Player.

Reguas:
PLAY_AMBIENT_SPEECH_FROM_POSITION_NATIVE plays multiple sound files assigned with the speech name alternately. In the target voice package there are files BUDDY_DOWN_01 and BUDDY_DOWN_02. "He's got Menendez!" line is located in the second file, while the first one contains the other line ("Alvarado's down!"). I wanted to play the Menendez line only, and there's no way to do this using only natives.

ChaosMod/Memory/Audio.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include "../Util/Logging.h"
+#include "Handle.h"
+#include "Memory.h"
+#include "Util/Natives.h"
```
pongo1231, line 6:
```suggestion
#include "Handle.h"
#include "Memory.h"
#include "Util/Logging.h"
#include "Util/Natives.h"
```

pongo1231 (approved):

PR #3398, opened 2023-03-12
Added "Hello, This Is Agatha" effect

Rylxnd:
The annoying phone call from Ms. Baker from GTA:O plays. 

[Video](https://discord.com/channels/785656433529716757/862742658039545857/1084563480423706724)

~~If you want to test the effect via the Debug Menu, you will have have to disable line 64 in `DebugMenu.cpp` temporarily.~~


OnlyRealNubs:
The phone is now opened via Script Global vars instead of inputs :smile:

Rylxnd:
Going to re-open once I remake this :)

PR #3432, opened 2023-04-23
Added "ESP" Effect

MoneyWasted:
This effect implements an ESP (Extra Sensory Perception), which is a common cheat in games that allows the player to see other players or NPCs behind solid geometry. This specific ESP draws the skeleton of pedestrians within a certain distance from the player and with a random line color.

There are some technical limitations, such as the ESP being drawn one frame behind the rest of the game. This is due to how Rockstar handles rendering. This can be solved by using other methods of rendering, such as hooking DirectX directly to draw the lines with a shader. This was my first approach as it would not cause this issue, but it became too much of a hassle.

The current method of how the line is rendered is by hooking the same function that is responsible for the other drawing natives, such as DRAW_RECT. Doing this allows me to reuse existing code that Rockstar has already developed.

Below is what the ESP looks like on a pedestrian.
Skeleton Drawing Example


ChaosMod/Effects/db/Player/MiscEsp.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+Vec2 GetBoneScreenCoords(Ped ped, int boneID)
+{
+	Vector3 boneCoords = GET_PED_BONE_COORDS(ped, boneID, 0, 0, 0);
+	Vec2 screenCoords;
+	Hooks::WorldToScreen(boneCoords, &screenCoords);
+	return screenCoords;
+}
+
+void DrawSkeleton(const std::array<Vec2, 19> &points, const std::array<std::array<int, 2>, 14> &connections,
```
pongo1231, line 39:
```suggestion
static void DrawSkeleton(const std::array<Vec2, 19> &points, const std::array<std::array<int, 2>, 14> &connections,
```

ChaosMod/Effects/db/Player/PlayerAutopilot.cpp
```diff
@@ -68,7 +68,7 @@ static void OnTick()
 
 #ifdef _DEBUG
 	// Draw debug go to line
-	if (m_state == STATE_TO_COORDS)
+	if (ms_State == STATE_TO_COORDS)
```
pongo1231, line 71: Whilst correct, this is out of the scope of the PR and should be done in a separate commit.

ChaosMod/Memory/Hooks/DrawingHooks.cpp
```diff
@@ -0,0 +1,76 @@
...
+// https://github.com/citizenfx/fivem/blob/master/code/components/extra-natives-five/src/Draw2dNatives.cpp
+
+// Notes:
+// Analysis in IDA shows that the return values of
+// _HK_allocateDrawRect is most likely a structure
+// of some sorts.
+
+static char *g_drawRects           = 0;
+static int *g_mainThreadFrameIndex = 0;
+static int g_drawRectsSize         = 0;
```
pongo1231, line 17:
```suggestion
static char *ms_DrawRects           = 0;
static int *ms_MainThreadFrameIndex = 0;
static int ms_DrawRectsSize         = 0;
```
(Of course has to be adapted in other places as well)

ChaosMod/Util/Types.h
```diff
@@ -2,6 +2,21 @@
...
 
+struct Vec2
+{
+	float x, y;
+};
+
+struct Vec3
+{
+	float x, y, z;
+};
```
pongo1231, line 13: Might be better to rename them to `ChaosVector2` and `ChaosVector3` to differentiate them from the existing Vector type (which adds padding between each member and is intended for interoperating with natives).

ChaosMod/Memory/Hooks/DrawingHooks.cpp
pongo1231, line 1: Any reason these (and the WorldToScreen ones) are hooks rather than just pointers to the functions in the game code? The hooks aren't doing anything other than wrapping the original functions. Might make more sense to just directly use them and move the code into ChaosMod/Memory/ instead.

ChaosMod/Effects/db/Player/PlayerAutopilot.cpp
MoneyWasted: This change is needed for compilation in debug mode. I will take it out once all issues are resolved and this pull request is ready for merging.

ChaosMod/Memory/Hooks/DrawingHooks.cpp
```diff
@@ -0,0 +1,83 @@
...
+	{
+		return false;
+	}
+
+	ms_DrawRects             = handle.At(0x32).Into().Get<char>();
+	ms_DrawRectsSize         = *handle.At(0x2C).Get<int>();
+	ms_MainThreadFrameIndex  = handle.At(0x0E).Into().Get<int>();
+
+	ms_AllocateDrawRectAddr  = handle.At(0x56).Into().Get<void *>();
+	ms_SetDrawRectCoordsAddr = handle.At(0x92).Into().Get<void *>();
```
pongo1231, line 39:
```suggestion
	ms_AllocateDrawRectAddr  = handle.At(0x56).Into().Get<void>();
	ms_SetDrawRectCoordsAddr = handle.At(0x92).Into().Get<void>();
```
No need for the double pointer

ChaosMod/Memory/Hooks/DrawingHooks.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "Util/Color.h"
+
+// Thanks CitizenFX!
+// Adapted by MoneyWasted for ChaosModV
+// https://github.com/citizenfx/fivem/blob/master/code/components/extra-natives-five/src/Draw2dNatives.cpp
+
+// This file is manually formatted.
+// clang-format off
+
+static const int MAX_LINES = 5000;
```
pongo1231, line 13: Nitpick but could be a macro instead to be consistent with the way the project defines constants

ChaosMod/Memory/Hooks/WorldToScreenHook.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
+
+namespace Hooks
+{
+	bool WorldToScreen(Vector3 worldPosition, rage::Vector2 *screenPosition);
```
pongo1231, line 5: Needs to include the corresponding headers for the types here

ChaosMod/Memory/Hooks/DrawingHooks.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+namespace Hooks
+{
+	void DrawLine(float x1, float y1, float x2, float y2, float width, int r, int g, int b, int a);
+
+	inline void DrawLine(rage::Vector2 firstPos, rage::Vector2 secondPos, Color color, float thickness)
```
pongo1231, line 7: Should also include the headers containing the types (or a forward declaration if you desire).

pongo1231 (changes requested):
The hooks still need to be moved up a folder and refactored to be in consistency with the other files in there. Also has the benefit of simplifying the code a bit by lazily evaluating the pattern and storing the result using a local static variable.

ChaosMod/ChaosMod.vcxproj
```diff
@@ -58,6 +58,7 @@
       <LanguageStandard>stdcpplatest</LanguageStandard>
       <PreprocessorDefinitions>_WINDLL;%(PreprocessorDefinitions);CURL_STATICLIB;NOMINMAX;_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS;_CRT_SECURE_NO_WARNINGS</PreprocessorDefinitions>
       <AdditionalIncludeDirectories>.;../vendor</AdditionalIncludeDirectories>
+      <AdditionalOptions>/bigobj %(AdditionalOptions)</AdditionalOptions>
```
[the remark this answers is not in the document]
MoneyWasted: Due to this error while compiling LuaScripts.cpp.
`C1128: number of sections exceeded object file format limit: compile with /bigobj`

ChaosMod/Memory/WorldToScreen.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include "Util/Types.h"
+
+bool (*OG_WorldToScreen)(ChaosVector3, float *, float *);
+
+static void *ms_WorldToScreenAddr = 0;
+
+namespace Hooks
```
pongo1231, line 9:
```suggestion
namespace Memory
```

ChaosMod/Memory/Drawing.h
```diff
@@ -0,0 +1,83 @@
...
+#include "Util/Color.h"
+
+// Thanks CitizenFX!
+// Adapted by MoneyWasted for ChaosModV
+// https://github.com/citizenfx/fivem/blob/master/code/components/extra-natives-five/src/Draw2dNatives.cpp
+
+// This file is manually formatted.
+// clang-format off
+
+constexpr int MAX_LINES = 5000;
```
pongo1231, line 12:
```suggestion
#define MAX_LINES 5000
```

ChaosMod/Memory/Drawing.h
```diff
@@ -0,0 +1,83 @@
...
+
+char *(*OG_AllocateDrawRect)(void *);
+void (*OG_SetDrawRectCoords)(void *, float, float, float, float);
```
pongo1231, line 15: No need for those, you can simply use local static variables (and drop the redundant OG_ prefix)

ChaosMod/Memory/WorldToScreen.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include "Util/Types.h"
+
+bool (*OG_WorldToScreen)(ChaosVector3, float *, float *);
```
pongo1231, line 5: Same here

ChaosMod/Memory/WorldToScreen.h
```diff
@@ -0,0 +1,40 @@
...
+
+static void *ms_WorldToScreenAddr = 0;
```
pongo1231, line 7: Duplicate

ChaosMod/Memory/Drawing.h
```diff
@@ -0,0 +1,83 @@
...
+namespace Hooks
+{
+	inline void DrawLine(float x1, float y1, float x2, float y2, float width, int r, int g, int b, int a)
+	{
+		static char *ms_DrawRects             = 0;
+		static int ms_DrawRectsSize           = 0;
+		static int *ms_MainThreadFrameIndex   = 0;
+
+		static void *ms_AllocateDrawRectAddr  = 0;
+		static void *ms_SetDrawRectCoordsAddr = 0;
```
pongo1231, line 26:
```suggestion
		static char *drawRects             = 0;
		static int drawRectsSize           = 0;
		static int *mainThreadFrameIndex   = 0;

		static void *allocateDrawRectAddr  = 0;
		static void *setDrawRectCoordsAddr = 0;
```
We don't do prefixes in this case. Again a nitpick but outside of effect files I have to be a bit more stingy for consistency purposes

ChaosMod/Memory/WorldToScreen.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Hooks
+{
+	inline bool WorldToScreen(Vector3 worldPosition, Vector3 *screenPosition)
+	{
+		static void *ms_WorldToScreenAddr = 0;
```
pongo1231, line 13:
```suggestion
		static void *worldToScreenAddr = 0;
```

ChaosMod/Memory/WorldToScreen.h
```diff
@@ -0,0 +1,40 @@
...
+
+			ms_WorldToScreenAddr = handle.Get<void>();
+			OG_WorldToScreen     = *(bool (*)(ChaosVector3, float *, float *))ms_WorldToScreenAddr;
+		}
+
+		ChaosVector3 _ {
+			.x = worldPosition.x,
+			.y = worldPosition.y,
+			.z = worldPosition.z,
+		};
```
pongo1231, line 31: The function should directly accept the ChaosVector3 rather than constructing one from another Vector3

pongo1231 (changes requested):

pongo1231:
The config utility entry for the effect is missing. Also can you test if the patterns work with multiple game builds to make sure they are build-agnostic enough? Otherwise this seems good enough.

MoneyWasted:
> The config utility entry for the effect is missing. Also can you test if the patterns work with multiple game builds to make sure they are build-agnostic enough? Otherwise this seems good enough.

* The world to screen pattern hasn't changed since I started using it on build 1604.
* I tested the pattern to the DRAW_RECT function on all the game builds I have access to, and it is the same on all of them.
* The pattern to 'cmp edx, 500' in AllocateDrawRect would only change if Rockstar changes the number of rectangles allowed to be drawn on the screen.

I have added the effect to the config app so this should be ready to merge!

ChaosMod/Effects/db/Player/MiscEsp.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+			DrawSkeleton(points, connections, lineColor, thickness);
+		}
+	}
+}
+
+REGISTER_EFFECT(OnStart, nullptr, OnTick, EffectInfo
+	{
+		.Name = "ESP",
+		.Id = "msic_esp",
```
pongo1231, line 97:
```suggestion
		.Id = "misc_esp",
```

ChaosMod/Effects/db/Player/MiscEsp.cpp
```diff
@@ -0,0 +1,101 @@
...
+		.IsTimed = true,
+		.EffectGroupType = EffectGroupType::Weapons
```
pongo1231, line 99: This doesn't modify weapons like the other effects so I don't see the need for this here

pongo1231 (changes requested):

ChaosMod/Effects/db/Player/MiscEsp.cpp
MoneyWasted: Would be helpful to have a "Visual" category. I know we have "Shaders" category, but some effects such as "No Sky" don't fit into that category. I will remove this for now.

PR #3441, opened 2023-04-29
Added an option to dispatch effects based on travelled distance

Reguas:
https://www.twitch.tv/videos/1803745281
https://www.twitch.tv/videos/1804685667
Meta effects are disabled in this mode; Twitch voting works fine


ChaosMod/Components/EffectDispatcher.h
```diff
@@ -64,6 +64,12 @@ class EffectDispatcher : public Component
 		}
 	};
 
+	enum TravelledDistanceType
```
pongo1231, line 67: Would be cleaner to use enum class here to not pollute the global namespace

ChaosMod/Components/EffectDispatcher.h
```diff
@@ -95,6 +101,13 @@ class EffectDispatcher : public Component
 	std::list<RegisteredEffect *> m_DispatchedEffectsLog;
 
   private:
+	bool m_EnableDistanceBasedEffectDispatch = false;
+
+	float m_DistanceToActivateEffect = 500.f;
+	Vector3 m_SavedPosition = Vector3();
```
pongo1231, line 107:
```suggestion
	Vector3 m_SavedPosition;
```

pongo1231 (changes requested):

ChaosMod/Components/EffectDispatcher.h
Reguas: 👌

ChaosMod/Components/EffectDispatcher.cpp
```diff
@@ -226,6 +226,11 @@ EffectDispatcher::EffectDispatcher(const std::array<BYTE, 3> &timerColor, const
 	m_MetaEffectShortDur =
 	    g_OptionsManager.GetConfigValue<int>("MetaShortEffectDur", OPTION_DEFAULT_EFFECT_META_SHORT_TIMED_DUR);
 
+	m_EnableDistanceBasedEffectDispatch = g_OptionsManager.GetConfigValue<bool>("EnableDistanceBasedEffectDispatch", false);
+	m_DistanceToActivateEffect = g_OptionsManager.GetConfigValue<int>("DistanceToActivateEffect", 250);
+	m_DistanceType                      = static_cast<TravelledDistanceType>(
+        g_OptionsManager.GetConfigValue<int>("DistanceType", static_cast<int>(TravelledDistanceType::Distance)));
```
pongo1231, line 232: The default value should be a def under Util/OptionsDefault.h for consistency with the other GetConfigValue calls

pongo1231 (changes requested):
Other than one thing LGTM

ChaosMod/Components/EffectDispatcher.cpp
Reguas: 👌

PR #3475, opened 2023-06-24
Fixed workshop not loading if it's tab wasn't selected by mouse

SRG27:
Changed the event that calls the function for loading workshop items from `MouseLeftButtonDown` to `Selector.Selected` so that the workshop items are loaded when the user is navigating with a keyboard.


pongo1231:
Thanks!

Issue #3466, opened 2023-06-17
[Effect Suggestion] tiny screen


davidman2:
**Name of effect**
Tiny/small screen/resolution
**Duration**
- [ ] Long
- [x] Short
- [ ] No duration

**Effect Description**

When the effect is enabled the screen will turn black but the actual gameplay will be shown in the middle with screen resolution so low that the gameplay just because tiny and most of the screen will just become black. To give an example imagine the dvd screen saver effect but the the rectangle that's bouncing won't be 180 degrees it will be 90 degrees and instead of bouncing around it will stay still in the middle of the screen and it will also be smaller.

PR #3479, opened 2023-07-01
Added "Tiny Screen" Effect

SRG27:
As suggested in #3466.

Reduces the screen dimensions by 9 to 40 times and centers it.


Rylxnd:
Seems to sound a lot like #3000 🧐

we-script:
@everyone

PR #3514, opened 2023-08-08
Added "Bouncy Radar"

SRG27:
Bounces the radar across the screen like the [DVD screensaver](https://www.youtube.com/watch?v=5mGuCdlCcNM).

### DIFF
--- a/ChaosMod/Effects/db/Screen/ScreenBouncyRadar.cpp
+++ b/ChaosMod/Effects/db/Screen/ScreenBouncyRadar.cpp
@@ -1,0 +1,56 @@
+#include <stdafx.h>
+#include "Memory/UI.h"
+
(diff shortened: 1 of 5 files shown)


pongo1231:
Make sure to run clang-format, there's some broken formatting in there. VS should usually detect the .clangformat file and automatically do it on saving.

SRG27:
Done 👍🏼

ChaosMod/Effects/db/Screen/ScreenBouncyRadar.cpp
```diff
@@ -0,0 +1,64 @@
+#include "Memory/UI.h"
+#include <stdafx.h>
```
pongo1231, line 2:
```suggestion
#include <stdafx.h>

#include "Memory/UI.h"
```
stdafx.h should always be the first included header in source files, you can add an empty line between both to stop clang-format from reshuffling those

Issue #3153, opened 2022-07-10
[Effect Suggestion] Mrs. Baker Calls You Again


WeirdPov:
**Name of effect**
Mrs. Baker Calls You Again

**Duration (choose one and replace the corresponding "[ ]" with "[X]")**
- [ ] Long
- [ ] Short
- [X] No duration

**Effect Description**
_"If its existed already, i'll remove this as planned"_
From DarkViperAU's video about agatha, on top comment says 
"I really want this to be an unskippable phone call effect in chaos."
By: TotalHydration (On top comment)
Video: https://www.youtube.com/watch?v=ttPmvX0LIKE

OnlyRealNubs:
Duplicate of [#Call Spam](https://github.com/gta-chaos-mod/ChaosModV/issues/1119) . For the love of god, look for duplicates.

WeirdPov:
> Duplicate of [#1119](https://github.com/gta-chaos-mod/ChaosModV/issues/1119) . For the love of god, look for duplicates.

Yea i know!! i put there as ""If its existed already, i'll remove this as planned"" GAD DAMMIT

OnlyRealNubs:
> > Duplicate of [#1119](https://github.com/gta-chaos-mod/ChaosModV/issues/1119) . For the love of god, look for duplicates.
> 
> Yea i know!! i put there as ""If its existed already, i'll remove this as planned"" GAD DAMMIT

Maybe check before hand we wouldn't be here. 🙃

Issue #3128, opened 2022-07-02
[Effect Suggestion] Mr.Baker Phone call


Troi4839:
**Name of effect**
Potential name(s) for the effect
Mr.Baker Call

**Duration (choose one and replace the corresponding "[ ]" with "[X]")**
- [ ] Long
- [ ] Short
- [X] No duration

**Effect Description**
Description of the effect; what should it do?
It should start the Agatha Baker phone call and it should be unskippable

Last0xygen:
Duplicate nr 6

Issue #3127, opened 2022-07-02
[Effect Suggestion] Add Agatha Baker's phone call


ParthNaik14:
**Name of effect**
Agatha strikes again

**Duration (choose one and replace the corresponding "[ ]" with "[X]")**
- [ ] Long
- [ ] Short
- [ ] No duration

**Effect Description**
This phone call is a running joke in whole of GTA 5 community. Sure, this became so famous as it is because of DarkViperAU, but it's not gonna cater to just one specific streamer's audience. The whole community in is annoyed by this phone call and makes fun of it. It'll be one of the best effects ever. There's no loss in adding the effect, it's a win-win situation. You anyways give the option to opt out of effects by manipulating the .ini file.

Last0xygen:
> It will become an instant meme.

It already is, isn’t it? ;)

> There's absolutely no reason why not to put this effect. 

You‘re more than welcome to open a PR and I’ll review it.
But I think it is prone to blocking mission / cutscenes, so that could be a problem

> It'll be one of the best effects ever

ok

Issue #3126, opened 2022-07-02
[Effect Suggestion] Hello, This is Agatha.


katnitro:
**Name of effect**
Hello, This is Agatha.
This is Agatha.
Mrs. Baker

**Duration (choose one and replace the corresponding "[ ]" with "[X]")**
- [ ] Long
- [ ] Short
- [x ] No duration

**Effect Description**
To celebrate matt finally ridding his GTA Online experience from the spam of Agatha's phone calls, we should reward him with a chaos effect that simulates the iconic phone call from Mrs Baker. This phone call would obviously be the classic phone call we've heard many times on Matto's Online streams but completely unskippable and unable to hang up on so he has no choice but to listen to the entire phone call.

Last0xygen:
Duplicate https://github.com/gta-chaos-mod/ChaosModV/issues/2902

PR #3528, opened 2023-08-17
Add effects "Hello, this is Agatha" & "Dave Here"

### LINKED ISSUE #3523: [Effect Improvement] Hello, this is Agatha
**Name of effect**
Improvement for the "Hello, this is Agatha" effect

**Duration**
- [ ] Long
- [ ] Short
- [X] No duration

**Effect Description**
Replace the effect name with "Annoying phone call" and randomly pick one unskippable famously annoying GTAO phone call. 
It could be English Dave (The "Knock knock !" thing), Agatha (with the casino), Ron (calling about the hangar), Moodyman (calling about his slippers), etc... It could be any phone call even one from the story mode, obviously the longer the better !

I love the Agatha effect so kudos to Rylxnd for making it, but imo this improvement would add a lot more fun to it :)

Rylxnd:
Fixes #3523, #2521, #2902, #3126, #3127, #3128 & #3153 😄 

<hr>

Once the effect(s) are triggered, the player will open their phone an receive a call. They are forced to take the call and can't hang up. (Don't worry the call isn't that long)

The effects should work with future patches, I finally figured out how to get the YSC globals. (Thanks Parik)

### DIFF
--- a/ChaosMod/Components/Globals.cpp
+++ b/ChaosMod/Components/Globals.cpp
@@ -1,0 +1,71 @@
+#include <stdafx.h>
+
+#include "Globals.h"
+
+#include "Lib/scrThread.h"
+
+#include "Memory/Script.h"
+
+std::map<std::string, int> Globals::m_GlobalsIndexMap = {};
+static std::vector<GlobalRegistration> m_GlobalsRegistration = {};
+
+Globals::Globals() : Component()
+{
+	REGISTER_GLOBAL("CellPhoneState", "dialogue_handler",
+	                "2D 00 02 00 ? 52 ? ? 41 ? 72 08 2A 06 56 ? ? 52 ? ? 41 ? 71 08 20 56 ? ? 72", 6,
+	                GlobalPatternIdiom::GLOBAL_U16);	
+	
+	REGISTER_GLOBAL("CellPhoneLock", "dialogue_handler",
+	                "72 54 ? ? 5d ? ? ? 71 2c ? ? ? 2b 72 54 ? ? 77 54 ? ? 52 ? ? 25 ? 2c ? ? ? 53 ? ? 06 56 ? ? 52 ? ? 76", 2,
+	                GlobalPatternIdiom::GLOBAL_U16);
+
+
+	m_SearchGlobalsListener.Register(
+	    Hooks::OnScriptThreadRun,
+	    [&](rage::scrThread *thread)
+	    {
+			// Save a f**k-tonne of CPU resources by destructing the event listener when we no longer have anymore globals to find
+			if (m_GlobalsRegistration.size() <= 0)
+			{
+			    m_SearchGlobalsListener.~ChaosEventListener();
+			    return true;
+			}
+
+		    auto match =
+		        std::find_if(m_GlobalsRegistration.begin(), m_GlobalsRegistration.end(),
+		                     [&thread](GlobalRegistration &reg) { return reg.m_ScriptName == thread->GetName(); });
+
+		    if (match == m_GlobalsRegistration.end())
+			    return true;
+
+		    GlobalRegistration searchedGlobal = *match;
+
+		    auto program = Memory::ScriptThreadToProgram(thread);
+		    if (program->m_CodeBlocks)
+		    {
+			    Handle handle = Memory::FindScriptPattern(searchedGlobal.m_Pattern, program);
+
+			    if (!handle.IsValid())
+			    {
+				    LOG("Failed to find global \"" << searchedGlobal.m_Name << "\"");
+			    }
+			    else
+			    {
+				    int globalIndex;
+
+				    if (searchedGlobal.m_PatternIdiom == GlobalPatternIdiom::GLOBAL_U16)
+					    globalIndex = handle.At(searchedGlobal.m_Offset).Value<uint16_t>() & 0xFFFFFF;
+				    else
+					    globalIndex = handle.At(searchedGlobal.m_Offset).Value<uint32_t>() & 0xFFFFFF;
+
+				    Globals::SetGlobalIndex(searchedGlobal.m_Name, globalIndex);
+
+				    LOG("Found global \"" << searchedGlobal.m_Name << "\" (Global: " << globalIndex << ")");
+			    }
+		    }
+
+		    m_GlobalsRegistration.erase(match);
+
+		    return true;
+	    });
+}
--- a/ChaosMod/Components/Globals.h
+++ b/ChaosMod/Components/Globals.h
@@ -1,0 +1,66 @@
+#pragma once
+
+#include "Components/Component.h"
+
+#include "Memory/Hooks/ScriptThreadRunHook.h"
+
+#include <map>
+#include <vector>
+
+enum GlobalPatternIdiom : uint8_t
+{
+	GLOBAL_U16,
+	GLOBAL_U24
+};
+
+struct GlobalRegistration
+{
+  public:
+	std::string m_Name;
+	std::string m_ScriptName;
+	std::string m_Pattern;
+	int m_Offset;
+	GlobalPatternIdiom m_PatternIdiom;
+
+	GlobalRegistration(std::string name, std::string scriptName, std::string pattern, int offset,
+	                   GlobalPatternIdiom patternIdiom)
+	    : m_Name(name), m_ScriptName(scriptName), m_Pattern(pattern), m_Offset(offset), m_PatternIdiom(patternIdiom)
+	{
+	}
+};
+
+class Globals : public Component
+{
+  private:
+	static std::map<std::string, int> m_GlobalsIndexMap;
+
+	CHAOS_EVENT_LISTENER(Hooks::OnScriptThreadRun) m_SearchGlobalsListener;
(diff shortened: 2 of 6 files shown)


ChaosMod/Components/Globals.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+Globals::Globals() : Component()
+{
+	REGISTER_GLOBAL("CellPhoneState", "dialogue_handler",
+	                "2D 00 02 00 ? 52 ? ? 41 ? 72 08 2A 06 56 ? ? 52 ? ? 41 ? 71 08 20 56 ? ? 72", 6,
+	                GlobalPatternIdiom::GLOBAL_U16);	
+	
+	REGISTER_GLOBAL("CellPhoneLock", "dialogue_handler",
+	                "72 54 ? ? 5d ? ? ? 71 2c ? ? ? 2b 72 54 ? ? 77 54 ? ? 52 ? ? 25 ? 2c ? ? ? 53 ? ? 06 56 ? ? 52 ? ? 76", 2,
+	                GlobalPatternIdiom::GLOBAL_U16);
```
pongo1231, line 20: Might be better to search for those in the effect directly I think? Otherwise this is just going to clutter up this file if we have too many of those. Also it's more consistent to the lazy search approach we have with patterns in the codebase.

pongo1231 (changes requested):

ChaosMod/Components/Globals.h
```diff
@@ -0,0 +1,76 @@
...
+	                   GlobalPatternIdiom patternIdiom)
+	    : m_Name(name), m_ScriptName(scriptName), m_Pattern(pattern), m_Offset(offset), m_PatternIdiom(patternIdiom)
+	{
+	}
+};
+
+class Globals : public Component
+{
+  private:
+	static inline std::map<std::string, int> m_GlobalsIndexMap = {};
```
pongo1231, line 35:
```suggestion
	static inline std::map<std::string, int> ms_GlobalsIndexMap = {};
```
(Also needs to be changed everywhere else)

ChaosMod/Components/Globals.cpp
```diff
@@ -0,0 +1,73 @@
+#include <stdafx.h>
+
+#include "Globals.h"
+
+#include "Lib/scrThread.h"
+
+#include "Memory/Script.h"
+
+std::vector<GlobalRegistration> m_GlobalsRegistration = {};
```
pongo1231, line 9: Was this meant to be a part of the class rather than global?

ChaosMod/Components/Globals.h
```diff
@@ -0,0 +1,76 @@
+#pragma once
+
+#include "Components/Component.h"
+
+#include "Memory/Hooks/ScriptThreadRunHook.h"
+
+#include <map>
+#include <vector>
+
+enum GlobalPatternIdiom : uint8_t
```
pongo1231, line 10:
```suggestion
enum class GlobalPatternIdiom : uint8_t
```
To not pollute the global namespace as much

ChaosMod/Effects/db/Misc/MiscPhoneCall.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+static int phoneScaleform                              = 0;
+static int currentMode                                 = 0;
+static int lastTick                                    = 0;
+static int soundId                                     = 0;
+
+static bool isDialing                                  = true;
+static bool lockCellphone                              = true;
+
+std::unordered_map<Hash, const char *> phoneScaleforms = { { 0xD7114C9, "cellphone_ifruit" },
```
pongo1231, line 45:
```suggestion
static std::unordered_map<Hash, const char *> phoneScaleforms = { { 0xD7114C9, "cellphone_ifruit" },
```

ChaosMod/Components/Globals.h
```diff
@@ -0,0 +1,76 @@
...
+	std::string m_Pattern;
+	int m_Offset;
+	GlobalPatternIdiom m_PatternIdiom;
+
+	GlobalRegistration(std::string name, std::string scriptName, std::string pattern, int offset,
+	                   GlobalPatternIdiom patternIdiom)
+	    : m_Name(name), m_ScriptName(scriptName), m_Pattern(pattern), m_Offset(offset), m_PatternIdiom(patternIdiom)
+	{
+	}
+};
```
pongo1231, line 30: Can be moved inside the Globals class definition

ChaosMod/Effects/db/Misc/MiscPhoneCall.cpp
```diff
@@ -0,0 +1,356 @@
...
+static void PlayPhoneCall(PhoneCall &phoneCall)
+{
+	REGISTER_GLOBAL("CellPhoneState", "dialogue_handler",
+	                "2D 00 02 00 ? 52 ? ? 41 ? 72 08 2A 06 56 ? ? 52 ? ? 41 ? 71 08 20 56 ? ? 72", 6,
+	                GlobalPatternIdiom::GLOBAL_U16);
+
+	REGISTER_GLOBAL(
+	    "CellPhoneLock", "dialogue_handler",
+	    "72 54 ? ? 5D ? ? ? 71 2C ? ? ? 2B 72 54 ? ? 77 54 ? ? 52 ? ? 25 ? 2C ? ? ? 53 ? ? 06 56 ? ? 52 ? ? 76", 2,
+	    GlobalPatternIdiom::GLOBAL_U16);
```
pongo1231, line 112: Since we are calling those on-demand now it might be better design-wise to instead return a handle object which can then be passed to GlobalExists and GetGlobalAddr. These handles can then be cached in the effect itself so we don't have to worry about implementing some caching mechanism in the component itself.

ChaosMod/Components/Globals.h
```diff
@@ -0,0 +1,76 @@
...
+	static void RegisterGlobal(std::string name, std::string scriptName, std::string pattern, int patternOffset,
+	                           GlobalPatternIdiom patternIdiom);
+
+	template <class T>
+	requires std::is_base_of_v<Component, T>
+	friend struct ComponentHolder;
+};
+
+#define REGISTER_GLOBAL(name, scriptName, pattern, patternOffset, patternIdiom) \
+	Globals::RegisterGlobal(name, scriptName, pattern, patternOffset, patternIdiom);
```
pongo1231, line 76: Actually, the macro seems redundant in this case? Unless it's supposed to provide some benefit I'm not seeing here.

ChaosMod/Effects/db/Misc/MiscPhoneCall.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+	while (!Globals::GlobalExists("CellPhoneState") || !Globals::GlobalExists("CellPhoneLock"))
+	{
+		WAIT(0);
+	}
+
+	lockCellphoneGlobal  = Globals::GetGlobalAddr<uint8_t>("CellPhoneLock");
+	cellphoneStateGlobal = Globals::GetGlobalAddr<CellPhone>("CellPhoneState");
```
pongo1231, line 120: There should be a way to check if those globals were actually found (a GlobalFound function or such) and a fallback if not so - in this case probably just making the effect do nothing for the full duration.

ChaosMod/Components/Globals.cpp
```diff
@@ -0,0 +1,73 @@
...
+				    LOG("Failed to find global \"" << searchedGlobal.m_Name << "\"");
+			    }
+			    else
+			    {
+				    int globalIndex;
+
+				    if (searchedGlobal.m_PatternIdiom == GlobalPatternIdiom::GLOBAL_U16)
+					    globalIndex = handle.At(searchedGlobal.m_Offset).Value<uint16_t>() & 0xFFFFFF;
+				    else
+					    globalIndex = handle.At(searchedGlobal.m_Offset).Value<uint32_t>() & 0xFFFFFF;
```
pongo1231, line 47:
```suggestion
				    if (searchedGlobal.m_PatternIdiom == GlobalPatternIdiom::GLOBAL_U16)
				    {
					    globalIndex = handle.At(searchedGlobal.m_Offset).Value<uint16_t>() & 0xFFFFFF;
				    }
				    else
				    {
					    globalIndex = handle.At(searchedGlobal.m_Offset).Value<uint32_t>() & 0xFFFFFF;
				    }
```

ChaosMod/Components/Globals.cpp
```diff
@@ -0,0 +1,73 @@
...
+		    {
+			    return true;
+		    }
+
+		    auto match =
+		        std::find_if(m_GlobalsRegistration.begin(), m_GlobalsRegistration.end(),
+		                     [&thread](GlobalRegistration &reg) { return reg.m_ScriptName == thread->GetName(); });
+
+		    if (match == m_GlobalsRegistration.end())
+			    return true;
```
pongo1231, line 27:
```suggestion
		    if (match == m_GlobalsRegistration.end())
		    {
			    return true;
		    }
```

pongo1231 (changes requested):
Was the issue with the long delay between the global registration and finding it resolved?

PR #3538, opened 2023-08-18
Expose "TeleportPlayer" to Lua Runtime

Rylxnd:


ChaosMod/Components/LuaScripts.cpp
```diff
@@ -599,6 +599,8 @@ LuaScripts::ParseScriptReturnReason LuaScripts::ParseScriptRaw(std::string scrip
 		E("GetAllPeds", GetAllPedsArray),
 		E("CreatePoolPed", CreatePoolPed),
 
+		E("TeleportPlayer", [](LuaVector3 pos, bool noOffset) { TeleportPlayer(pos.X, pos.Y, pos.Z, noOffset); }),
```
pongo1231, line 602: Would this need to be called like `TeleportPlayer(LuaVector3(1.0, 1.0, 1.0), false);`? If so it might be better to just directly take the x,y,z as params.

ChaosMod/Components/LuaScripts.cpp
```diff
@@ -599,6 +599,8 @@ LuaScripts::ParseScriptReturnReason LuaScripts::ParseScriptRaw(std::string scrip
 		E("GetAllPeds", GetAllPedsArray),
 		E("CreatePoolPed", CreatePoolPed),
 
+		E("TeleportPlayer", [](float x, float y, float z, bool noOffset) { TeleportPlayer(x, y, z, noOffset); }),
```
pongo1231, line 602:
```suggestion
		E("TeleportPlayer", TeleportPlayer),
```
No reason to wrap it with the previous changes

pongo1231 (approved):

pongo1231:
I had assumed `TeleportPlayer` to be taking x, y and z coordinates directly when it's actually a Vector3. The wrapper around the function was correct.

PR #3555, opened 2023-08-29
Fix crash on "Black Hole"

### LINKED ISSUE #3554: [Bug] "Black Hole" effect causes game to crash.
**Describe the bug**
Whenever the "Black Hole" effect is spawned, the game crashes.

**To Reproduce**
I did some testing by choosing the effect manually.

Rylxnd:
Fixes #3554


pongo1231:
Any way to semi-consistently replicate this crash? I can't reproduce it on my end and the fix doesn't particularly make sense to me either as the list shouldn't contain any invalid entities to begin with.

Rylxnd:
I'm not sure the cause. It crashed on my end every time. And adding the entity check seemed to resolve it both for me and other people.

pongo1231:
Since it apparently fixes that issue on certain setups and doesn't affect anything else I'll merge this. Worth checking if the crashing persists with other effects affecting entities though, especially ones which affect everything as Black Hole does.

PR #3556, opened 2023-08-29
Expose "CreateHostilePed" to Lua Runtime

Rylxnd:


Reguas:
Why not use [https://github.com/gta-chaos-mod/ChaosModV/blob/406031f148b60e1862d2b64b6d1cdff90fb7f0c6/ChaosMod/Util/Peds.h#L7](https://github.com/gta-chaos-mod/ChaosModV/blob/406031f148b60e1862d2b64b6d1cdff90fb7f0c6/ChaosMod/Util/Peds.h#L7)? 🤔

Rylxnd:
You got me f'd up. I looked for that but couldn't find it lol. I just assumed it got removed

pongo1231:
That won't work as that function takes a different Vector3 object to the one used by the Lua runtime (which is also why it was removed [before](https://github.com/gta-chaos-mod/ChaosModV/commit/a9a55780b21cc0b3102386023f3d0e90244b910a)).

Rylxnd:
I'll have to wrap it then, like TeleportPlayer

Rylxnd:
It would make it a lot easier for LUA. Since you don't have to type out all natives and can just use the built in function.

pongo1231:
Any progress with this?

PR #3564, opened 2023-09-05
Added Real First Person effect

Chanka0:
Forced First Person but EVEN MORE ~~painful~~ immersive. Vanilla first person prioritizes the player's mouse movements and moves the camera accordingly, this effect does the opposite. While active the player's view is changed to a camera that is driven by the player character's head movements. Aiming has been adjusted for usability.

If you want an example of it in use look at [footage from the GTA V first person cutscenes mod](https://www.youtube.com/watch?v=BFUf048B30o). The main difference between the two is camera roll being enabled which can be adjusted by replacing `boneRot.y` with `0` in `line 78`. Also the effect is not currently enabled in cutscenes but that can be easily remedied.

`Edits: Grammar, name change`


ChaosMod/Effects/db/Player/PlayerGoPro.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+static void OnStart()
+{
+	//Not sure of a direct way to obtain bone rotation. Attaching a prop is a workaround.
+	Vector3 boneCoords = GET_PED_BONE_COORDS(PLAYER_PED_ID(), 0x322c, 0, 0, 0);
+	goPro = CREATE_CAM_WITH_PARAMS("DEFAULT_SCRIPTED_CAMERA", boneCoords.x, boneCoords.y, boneCoords.z, 0, 0, 0, 60, false, 2);
+	boneProxy = CREATE_OBJECT("p_cs_leaf_s"_hash, boneCoords.x, boneCoords.y, boneCoords.z, false, false, true);
+	SET_ENTITY_ALPHA(boneProxy, 0, false);
+	ATTACH_ENTITY_TO_ENTITY(boneProxy, PLAYER_PED_ID(), GET_PED_BONE_INDEX(PLAYER_PED_ID(), 0x322c), 0.1, 0, 0, 0, 90,
```
Rylxnd, line 29: I would cache the player ped in a variable to reduce the amount of times you have the call for it.

ChaosMod/Effects/db/Player/PlayerGoPro.cpp
```diff
@@ -0,0 +1,99 @@
+/*
+    Effect by Chanka
+*/
+
+#include <stdafx.h>
+
+Cam goPro;
```
Rylxnd, line 7: Along with these vars being static

Rylxnd (changes requested):

Chanka0:
Pushed commit with suggested changes

ChaosMod/Effects/db/Player/PlayerGoPro.cpp
```diff
@@ -0,0 +1,100 @@
...
+static void OnTick()
+{
+	//SafetyCheck() checks for potential situations where the camera could become stuck and the switch statement below handles things if SafetyCheck is triggered.
+	switch (state)
+	{
+		//Normal operation
+		case 0:
+			if (SafetyCheck())
+			{
+			    OnStop();
```
Rylxnd, line 62: Best practice would be to avoid Calling the OnStart/OnStop functions. If you need to re-use the logic, move it to a new function and call that instead.

Rylxnd (changes requested):
This should also be moved with the rest of the Camera effects in the Screen category.

Chanka0:
Changes have been made and tested. Also it's worth mentioning that there are camera effects in the player category, such as GTA2, LagCamera, and ZoomZoomCam.

Reguas:
> there are camera effects in the player category, such as GTA2, LagCamera, and ZoomZoomCam.

Yeah, they need to be moved to the Screen category

pongo1231:
Haven't been able to test it yet but if it's basically a more extreme version of the First Person effect we could just straight up replace it with this version. Unless there are some caveats with it.

Also I'd be for enabling cutscene support, given it doesn't break anything (e.g. the camera isn't in a meddled state if the effect ends during a cutscene).

Chanka0:
The only caveat of this effect is potential motion sickness which is
already a thing thanks to Spinning Camera. Aiming was intentionally made
smoother for when the player was driving/running/etc.

Also cutscene support isn't an issue. There's a specific function to
enable/disable cutscene cameras taking priority so as soon as the effect
ends the cutscene camera can take over instantly. I wasn't about including
it because it's like watching a movie edited together from the director
assistants camera. You see all the inconsistencies and spot lighting change
between cuts.

pongo1231:
> The only caveat of this effect is potential motion sickness which is
already a thing thanks to Spinning Camera. Aiming was intentionally made
smoother for when the player was driving/running/etc.

In that case I think we should go with it and replace the existing First Person effect with this version rather than having 2 same-ish effects with one just being a more extreme version of the other.

>Also cutscene support isn't an issue. There's a specific function to
enable/disable cutscene cameras taking priority so as soon as the effect
ends the cutscene camera can take over instantly. I wasn't about including
it because it's like watching a movie edited together from the director
assistants camera. You see all the inconsistencies and spot lighting change
between cuts.

Seems perfectly fine to me. That'd just add to the chaos and breaking the game in weird ways is part of the fun of the mod.

Chanka0:
Alright, I'll add in the cutscene support soon. btw if the first person effect is being replaced, should there be something done to signify that it's new/changed?

pongo1231:
No other action is strictly necessary. Renaming the effect to "Real First Person" seems like a neat idea though and would indeed make players aware something has changed with the effect.

MoneyWasted:
for legal reasons changing the name of the effect would be a good move, how about “Action Mount”

Chanka0:
If this is replacing First Person I think Real First Person works best. It
reflects what the effect is while showing that it is new and different from
what it's replacing.

Unless anyone disagrees with it I'll change the name to that and remove the
old First Person

pongo1231:
No disagreement here 👍

ChaosMod/Effects/db/Screen/ScreenRealFirstPerson.cpp
```diff
@@ -0,0 +1,116 @@
...
+	if (DOES_CAM_EXIST(eCamera))
+	{
+		SET_CAM_ACTIVE(eCamera, false);
+		DESTROY_CAM(eCamera, true);
+	}
+}
+
+static void SetupCamera()
+{
+	// Not sure of a direct way to obtain bone rotation. Attaching a prop is a workaround.
```
Reguas, line 41: `ENTITY::GET_ENTITY_BONE_ROTATION(..., PED::GET_PED_BONE_INDEX(...))` 👍

Chanka0: Implemented it.

PR #3569, opened 2023-09-12
New effect: Cocktail shaker

Veloscocity:
Shakes all vehicles and props (not peds) back and forth, perpendicular to the direction the player is facing when the effect is triggered.


MoneyWasted (changes requested):
Ensure that you are using clang formatting on `MiscCocktail.cpp`. The easiest way to achieve this in Visual Studio is by pressing `CTRL+K` followed by `CTRL+D`. This will format the current file. Otherwise, everything looks fine. I'll let Pongo do the final review.

ChaosMod/Effects/db/Misc/MiscCocktail.cpp
```diff
@@ -0,0 +1,49 @@
...
+		Memory::ApplyForceToEntityCenterOfMass(veh, 1, y, x, 0, false, false, true, false);
+}
+
+// clang-format off
+REGISTER_EFFECT(OnStart, nullptr, OnTick, EffectInfo
+	{
+		.Name = "Cocktail Shaker",
+		.Id = "cocktail_shaker",
+		.IsTimed = true,
+        .IsShortDuration = true,
```
pongo1231, line 46:
```suggestion
		.IsShortDuration = true,
```
Formatting is broken

ChaosMod/Effects/db/Misc/MiscCocktail.cpp
```diff
@@ -0,0 +1,49 @@
...
+		lastTick = curTick;
+		x        = -x;
+		y        = -y;
+	}
+
+	for (auto object : GetAllProps())
+		Memory::ApplyForceToEntityCenterOfMass(object, 1, y, x, 0, false, false, true, false);
+
+	for (auto veh : GetAllVehs())
+		Memory::ApplyForceToEntityCenterOfMass(veh, 1, y, x, 0, false, false, true, false);
```
pongo1231, line 37:
```suggestion
	for (auto object : GetAllProps())
	{
		Memory::ApplyForceToEntityCenterOfMass(object, 1, y, x, 0, false, false, true, false);
	}

	for (auto veh : GetAllVehs())
	{
		Memory::ApplyForceToEntityCenterOfMass(veh, 1, y, x, 0, false, false, true, false);
	}
```
Usually not too strict with the code style in effects but this should be consistent with the rest of the project

ChaosMod/Effects/db/Misc/MiscCocktail.cpp
```diff
@@ -0,0 +1,49 @@
+//
+// effect by veloscocity
+//
+
+#include "Memory/Gravity.h"
+#include <math.h>
+#include <stdafx.h>
```
pongo1231, line 7:
```suggestion
#include <stdafx.h>

#include "Memory/Gravity.h"
```
`stdafx.h` needs to be the first include to not cause issues. Also `math.h` should be added there instead.

ChaosMod/Effects/db/Misc/MiscCocktail.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+static DWORD lastTick = 0;
+static void OnStart()
+{
+	Ped playerPed = PLAYER_PED_ID();
+	float rot =
+	    GET_ENTITY_HEADING(IS_PED_IN_ANY_VEHICLE(playerPed, false) ? GET_VEHICLE_PED_IS_IN(playerPed, false)
+	                                                               : playerPed); // Shake perpendicular to the player
+	x = sin((360 - rot) * PI / 180) * 1.33;
+	y = -cos((360 - rot) * PI / 180) * 1.33;
```
pongo1231, line 21:
```suggestion
	x = sin((360 - rot) * PI / 180) * 1.33f;
	y = -cos((360 - rot) * PI / 180) * 1.33f;
```
Don't think we need doubles here

pongo1231 (changes requested):

ChaosMod/Effects/db/Misc/MiscCocktail.cpp
MoneyWasted: precomputing pi/180 would save some CPU cycles.

ShufflePerson: PI us defined in multiple other effects, could it be worth adding it to a common header file? Of course PI won't change, though it would still feel ""cleaner"" to have it defined in one place.

Reguas: I believe it's better to use SYSTEM::SIN and SYSTEM::COS natives: they work with angles in degrees, so you won't have to use the value of pi.

ShufflePerson: `math.h` of DirectX has all of that. But I don't see one for PI/180

Value definitions: https://learn.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170#remarks

MoneyWasted: we could (and probably should) just create our own header file that has definitions for all the possible numbers we could need and just include that file in stdafx.h.

Veloscocity: so what am i supposed to do now

ConfigApp/Effects.cs
```diff
@@ -410,6 +410,7 @@ public enum EffectTimedType
             { "misc_fakeuturn", new EffectInfo("Fake U-Turn", EffectCategory.Misc) },
             { "misc_esp", new EffectInfo("ESP", EffectCategory.Misc, true) },
             { "screen_bouncyradar", new EffectInfo("Bouncy Radar", EffectCategory.Screen, true) },
+            { "cocktail_shaker", new EffectInfo("Cocktail Shaker", EffectCategory.Misc, true) },
```
pongo1231, line 413:
```suggestion
            { "cocktail_shaker", new EffectInfo("Cocktail Shaker", EffectCategory.Misc, true, true) },
```
Last flag is for whether the effect is short timed

pongo1231:
Looks good to me 👍

PR #3571, opened 2023-09-13
New Effect: Trip Everyone

Veloscocity:
Ragdolls all peds that aren't in a car and pushes them forward


Rylxnd:
Seems to similar to “Ragdoll Everyone” imo

Veloscocity:
I thought that too when I was finishing making it. I wanted to make basically a version of ragdoll everyone that was more likely to kill the player, but I think the tripping force is too weak for that. Maybe itll be better if I make the tripping force alot stronger

PR #3583, opened 2023-09-19
New Effect: Cyclone

Veloscocity:
When anything gets near the player, it gets spun around the player


ReiDaTecnologia:
Aren't that similar to Gravity Sphere?

Veloscocity:
Gravity sphere pulls things in directly and holds them in place relative to the player. Cyclone would spiral away anything close to the player.

ReiDaTecnologia:
Hopefully it is unique enough for them to accept then

PR #3595, opened 2023-10-05
New Effect: Six Wanted Stars

Veloscocity:
In addition to the police attention from 5 stars, the military will also hunt down the player in trucks, helicopters, boats, and tanks.


pongo1231:
Might make sense to offload it into its own file considering the amount of logic added for the effect

Veloscocity:
done

PR #3620, opened 2023-10-18
Add Model Spawn Bypass

MoneyWasted:
Allows all models to be spawned such as LOD models. Allows effects such as Reguas' "Inception" to work.

Example:


ReiDaTecnologia:
Modder!

PR #3652, opened 2024-02-15
Increase Wheel types to 12

holysnipz:
and added "random Wheel types" to Vehicle Rain effect as well

------

only tested in 3095. but should be fine with all versions, since GTA will default to Sport(0) if `WheelType` is *invalid* anyway.
*example: trying to apply Track(12) type in pre Tuners update* 

*p.s. and just noticed that my name in `credits.txt` is missing `h`* 😉


pongo1231:
Thanks!

> p.s. and just noticed that my name in credits.txt is missing h 😉

Sorry, got that fixed now.

PR #3653, opened 2024-02-17
future proof(?) shotgun checking logic

holysnipz:
no more hardcoded weapon hashes ~~*(even this is used by only 2 effects) 😅*~~

from SHV's natives definition, seems old enough to not cause compatibility issues in earlier game versions
`NATIVE_DECL Hash GET_WEAPONTYPE_GROUP(Hash weaponHash) { return invoke<Hash>(0xC3287EE3050FB74C, weaponHash); } // 0xC3287EE3050FB74C 0x5F2DE833 b323`

### DIFF
--- a/ChaosMod/Util/Weapon.h
+++ b/ChaosMod/Util/Weapon.h
@@ -17,6 +17,8 @@ namespace Util
 		case 984333226:
 		case -275439685:
 		case 317205821:
+		case 1432025498:
+		case 94989220:
 			return true;
 		}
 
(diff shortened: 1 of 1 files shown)


ShufflePerson:
> Okay, how does this perform compared to the “old” method with the switch case? Probably worse considering the speed of switch compared with the extra overhead of calling a native function.

Maybe adding caching could be the best option of both worlds? 
First call with each ID will be slower, the rest will be as fast as it was with switch.

pongo1231:
The overhead from the native call is very negligible, especially considering the already existing frequency of native calls by the mod. The additional optimization is appreciated regardless.

PR #3656, opened 2024-02-19
Real First Person: Added distance check

Chanka0:
Added a fix to the Real First Person effect. There was a bug where mission fails unrelated to player death/non-existence caused the camera to remain far away from the player. I've found the easiest fix is a simple distance check which covers any outlier situations as well as the affected mission fails.


pongo1231:
Thanks

PR #3693, opened 2025-01-21
Fixed "Flip UI" effect ID

Regynate:


pongo1231:
This will reset people's custom settings for that effect. We shouldn't touch it unless there's a good reason to do so.

Regynate:
True.

PR #3696, opened 2025-01-21
fixed some oversight(?) in recent fake tp pr

holysnipz:
this should be a review in #3692 instead if i didn't take a nap 😅


pongo1231:
Thanks

PR #3698, opened 2025-01-23
Added effects "Majority Voting" and "Minority Voting"

Regynate:


ChaosMod/Components/EffectDispatcher.cpp
```diff
@@ -409,13 +409,13 @@ void EffectDispatcher::UpdateMetaEffects(int deltaTime)
 		std::vector<std::tuple<EffectIdentifier, EffectData *>> availableMetaEffects;
 
 		float totalWeight = 0.f;
-		for (auto &[effectId, effectData] : g_EnabledEffects)
+		for (auto effectData : GetFilteredEnabledEffects())
```
pongo1231, line 412:
```suggestion
		for (auto &effectData : GetFilteredEnabledEffects())
```
So this doesn't potentially bite us in the future should GetFilteredEnabledEffects not return references anymore for whatever reason.

ChaosMod/Util/VotingMode.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include <string>
+
+class VotingMode
+{
+  public:
+	enum Value
```
pongo1231, line 8:
```suggestion
	enum class Value
```

pongo1231 (changes requested):

PR #3699, opened 2025-01-23
Added Crossing Challenge component

Regynate:
Allows to set starting and ending point; controls respawning at the start on death and shows a congratulatory message upon reaching the end. Also fixed a few bugs in the process.

### DIFF
--- a/ChaosMod/Components/EffectDispatcher.cpp
+++ b/ChaosMod/Components/EffectDispatcher.cpp
@@ -1,5 +1,6 @@
 #include <stdafx.h>
 
+#include "Components/CrossingChallenge.h"
 #include "Components/EffectDispatchTimer.h"
 #include "Components/EffectDispatcher.h"
 #include "Components/EffectSound/EffectSoundManager.h"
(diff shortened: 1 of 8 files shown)


ChaosMod/Components/EffectDispatcher.cpp
```diff
@@ -525,6 +526,8 @@ void EffectDispatcher::DispatchEffect(const EffectIdentifier &effectId, Dispatch
                                       const std::string &suffix)
 {
 	EffectDispatchQueue.push({ .Id = effectId, .Suffix = suffix, .Flags = dispatchEffectFlags });
+	if (ComponentExists<CrossingChallenge>())
+		GetComponent<CrossingChallenge>()->IncrementEffects();
```
pongo1231, line 530: Better to declare and fire a ChaosEvent for this and register a listener in the CrossingChallenge component instead (see this component and DebugSocket as reference).

ChaosMod/Util/OptionsFile.h
```diff
@@ -93,4 +114,20 @@ class OptionsFile
...
+		}
+		else
+		{
+			m_Options.emplace(szKey, value);
+		}
+	}
+	template <typename T> inline void SetValue(const std::string& szKey, T value)
+	{
+		SetValueString(szKey, std::to_string(value));
+	}
```
pongo1231, line 132:
```suggestion
	inline void SetValueString(const std::string& key, const std::string& value)
	{
		if (m_Options.contains((key))
		{
			m_Options[key] = value;
		}
		else
		{
			m_Options.emplace(key, value);
		}
	}
	template <typename T> inline void SetValue(const std::string& key, T value)
	{
		SetValueString(key, std::to_string(value));
	}
```
Hungarian notation bad

pongo1231 (changes requested):

PR #3703, opened 2025-01-27
Improvements to "Spawn Companion" effects

Regynate:
- Extracted common code to set relationships
- Made spawned peds like / hate the same entities player does

### DIFF
--- a/ChaosMod/Util/Peds.h
+++ b/ChaosMod/Util/Peds.h
@@ -56,11 +56,11 @@ inline Ped CreateHostilePed(Hash modelHash, Hash weaponHash, Vector3 *location =
 	return ped;
 }
 
-inline void SetCompanionRelationship(Ped ped, const char* name)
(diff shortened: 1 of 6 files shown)


ChaosMod/Util/Peds.h
```diff
@@ -55,3 +55,24 @@ inline Ped CreateHostilePed(Hash modelHash, Hash weaponHash, Vector3 *location =
 	SET_PED_FIRING_PATTERN(ped, 0xC6EE6B4C);
 	return ped;
 }
+
+inline void SetCompanionRelationship(Ped ped, const char* name)
```
pongo1231, line 59: Considering we expose it to the Lua runtime we should accept a `const std::string&` instead

pongo1231 (changes requested):

ChaosMod/Util/Peds.h
Regynate: Noted, didn't know that

PR #3705, opened 2025-01-28
Debug menu: Grey out filtered effects and display filter reason

Regynate:


ChaosMod/Effects/Condition/EffectCondition.h
```diff
@@ -3,17 +3,20 @@
...
 	VotingEnabled, // Voting is enabled
 	ProportionalVotingEnabled
 };
 
-#define REGISTER_EFFECT_CONDITION(conditionType, condition)                           \
-	namespace                                                                         \
-	{                                                                                 \
-		EffectCondition CONCAT(_effectCondition, __LINE__)(conditionType, condition); \
+#define REGISTER_EFFECT_CONDITION(conditionType, condition, description)                           \
+	namespace                                                                                      \
```
pongo1231, line 17: IMO "description" is non-descriptive, should rather be called "failReason" or something of the sort

pongo1231 (changes requested):

ChaosMod/Effects/Condition/EffectCondition.h
Regynate: Sadly I am very bad at naming things.

PR #3726, opened 2025-03-16
add buttons to discord bot

PalTheTaur:
Sorry that its sloppy, and I'd understand if this gets denied for that reason but it is something simple to make the discord bot so much nicer to use.


PalTheTaur:
sorry if this wasn't the way to go about this, I am still slightly unsure if I did this right.

PR #3741, opened 2026-02-01
fix: Discord voting race condition - add retry loop for guild connection

ghost:
- Add retry loop (up to 10 seconds) for guild.IsConnected check
- Explicitly add Guilds intent for proper guild data sync
- Add warning logs before handleFatal() calls for better debugging

Fixes issue where Discord voting fails immediately on startup because guild.IsConnected returns false right after Ready event fires.

### DIFF
--- a/TwitchChatVotingProxy/VotingReceiver/DiscordVotingReceiver.cs
+++ b/TwitchChatVotingProxy/VotingReceiver/DiscordVotingReceiver.cs
@@ -45,9 +45,9 @@ namespace TwitchChatVotingProxy.VotingReceiver
 
             m_Client = new DiscordSocketClient(new DiscordSocketConfig()
             {
-                // We don't actually need any but the Ready event fires much later without this for whatever reason
-                // So we just enable all unprivileged intents minus the ones the log warns us against using
-                GatewayIntents = GatewayIntents.AllUnprivileged
+                // Enable Guilds intent to properly receive guild data
+                // Also enable all unprivileged intents for proper functionality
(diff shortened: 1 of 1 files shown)


TwitchChatVotingProxy/VotingReceiver/DiscordVotingReceiver.cs
```diff
@@ -45,9 +45,9 @@ public async Task<bool> Init()
 
             m_Client = new DiscordSocketClient(new DiscordSocketConfig()
             {
-                // We don't actually need any but the Ready event fires much later without this for whatever reason
-                // So we just enable all unprivileged intents minus the ones the log warns us against using
-                GatewayIntents = GatewayIntents.AllUnprivileged
+                // Enable Guilds intent to properly receive guild data
+                // Also enable all unprivileged intents for proper functionality
+                GatewayIntents = GatewayIntents.Guilds | GatewayIntents.AllUnprivileged
```
holysnipz: redundant, no? since `Guilds` already part of `AllUnprivileged`
https://docs.discordnet.dev/api/Discord.GatewayIntents.html#fields

IcyNova420: While strictly redundant via bitwise definition, I'm keeping it explicit for readability and to ensure zero ambiguity given the flakiness of the Ready event documented in the comments above. The primary fix here is the retry loop.

Regynate:
LGTM 👍

